PR #23, opened 2020-12-17
[doc] use ddp to calculate loss of dev set

whiteshirt0429:


wenet/utils/executor.py
```diff
@@ -84,4 +84,11 @@ def cv(self, model, data_loader, device, args):
                                       batch_idx, num_total_batch, loss.item(),
                                       loss_att.item(), loss_ctc.item(),
                                       total_loss / num_seen_utts))
-        return total_loss / num_seen_utts
+
+        num_seen_utts = torch.Tensor([num_seen_utts]).to(loss.get_device())
+        # the default operator in all_reduce function is sum. 
+        average_params(num_seen_utts)
```
robin1001, line 90: directly calling torch.distributed.all_reduce here is better. 
There is no need to wrap it into a new function. And the function name avarage_params is ambiguous.

robin1001: And do this calling affect no DDP training?

robin1001 (changes requested):
please see inline

robin1001 (approved):

PR #57, opened 2021-01-14
[fix] Relative positional embedding

placebokkk:


placebokkk:
Explain the relative positional embedding in asr fully-attention encoder here:
http://placebokkk.github.io/asr/2021/01/14/asr-rpe.html

Espnet discuss ref: https://github.com/espnet/espnet/issues/2684


Opentransformer fix ref: https://github.com/ZhengkunTian/OpenTransformer/pull/28/files
Current fix seems not working for the gather method needs the index tenser having same shape with the input.

PR #93, opened 2021-01-26
Support left chunk size control.

TeaPoly:


robin1001:
Thanks for your contribution. The controlled left chunk size feature requires both training, decoding, and runtime modification, we have to do further experiments to verify the performance, and we will enable the feature in both training and runtime if it works well.

TeaPoly:
> Thanks for your contribution. The controlled left chunk size feature requires both training, decoding, and runtime modification, we have to do further experiments to verify the performance, and we will enable the feature in both training and runtime if it works well.

Okay.

robin1001:
Supported in another PR.

PR #94, opened 2021-01-26
[runtime] add compute_rtf for Android

pengzhendong:
Now I use cxxopts rather than gflags to parse the command line arguments.
If there is no problem, I will add the README later. 
Please tell me where I should place the README file.


runtime/server/x86/decoder/torch_asr_model.cc
```diff
@@ -16,7 +16,7 @@ void TorchAsrModel::Read(const std::string& model_path) {
   torch::jit::script::Module model = torch::jit::load(model_path);
   module_.reset(new TorchModule(std::move(model)));
   // For multi-thread performance
-  at::set_num_threads(1);
+  at::set_num_threads(4);
```
robin1001, line 19: for x86, it should be 1.  make it a parameter.

runtime/server/x86/bin/compute_rtf.cpp
```diff
@@ -0,0 +1,83 @@
...
+// limitations under the License.
+#include <chrono>
+#include <string>
+
+#include "decoder/symbol_table.h"
+#include "decoder/torch_asr_decoder.h"
+#include "decoder/torch_asr_model.h"
+#include "frontend/feature_pipeline.h"
+#include "frontend/wav.h"
+#include "glog/cxxopts.h"
```
robin1001, line 22: The code should work for both x86 and android, please use gflags

runtime/server/x86/bin/compute_rtf.cpp
```diff
@@ -0,0 +1,83 @@
...
+                                   decode_config);
+
+    int dur = wav_reader.num_sample() / 16;
+    auto start = chrono::steady_clock::now();
+    while (true) {
+      bool finished = decoder.Decode();
+      if (finished) {
+        break;
+      }
+    }
```
robin1001, line 73: it's better if we add a function in torch_asr_decoder,  which consumes all the input.

robin1001 (changes requested):
please see inline

runtime/device/android/wenet/app/src/main/cpp/CMakeLists.txt
```diff
@@ -5,38 +5,33 @@ set(CMAKE_CXX_STANDARD 14)
...
+    decoder/ctc_prefix_beam_search.cc
+    decoder/torch_asr_decoder.cc
+    decoder/torch_asr_model.cc)
+target_link_libraries(decoder PUBLIC log ${PYTORCH_LIBRARY} ${FBJNI_LIBRARY})
 
-target_include_directories(${TARGET} PRIVATE
-    ${PYTORCH_INCLUDE_DIRS}
-    ${PYTORCH_INCLUDE_DIRS}/torch/csrc/api/include
-    )
+link_libraries(frontend decoder android)
```
jschenxiaoyu, line 34: what is android here?

jschenxiaoyu, line 34: what is android here?

jschenxiaoyu, line 34: what is android here?

runtime/server/x86/bin/compute_rtf.cpp
```diff
@@ -0,0 +1,83 @@
+// Copyright (c) 2021 Mobvoi Inc (authors: Zhendong Peng)
```
jschenxiaoyu, line 1: also add compute_rtf binary into server cmake file and bazel build?

runtime/device/android/wenet/app/src/main/cpp/CMakeLists.txt
pengzhendong: It means link libraries to all following targets.

pengzhendong: libandroid_jni is needed by libpython_jni.

PR #98, opened 2021-01-26
Run compute_cmvn_stats.py only cmvn=true.

Honghe:


examples/aishell/s0/run.sh
```diff
@@ -60,9 +60,11 @@ if [ ${stage} -le 1 ] && [ ${stop_stage} -ge 1 ]; then
         cp -r data/$x $feat_dir
     done
 
-    tools/compute_cmvn_stats.py --num_workers 16 --train_config $train_config \
-        --in_scp data/${train_set}/wav.scp \
-        --out_cmvn $feat_dir/$train_set/global_cmvn
+    if [ "$cmvn" = true ]; then
```
robin1001, line 63: Here we designed it on purpose, we always compute CMVN stats on the data prepare stage, and it is easy to change CMVN in the following stage, otherwise we should redo the data compare once CMVN configure is changed.

robin1001 (commented):
please see inline.

PR #105, opened 2021-01-27
[fix] fix speed perturb and dither related bugs

placebokkk:
1. Use un-normalized dither value for  [-32768,32767] waveform
2. convert speed perturbed waveform scaled in [-1,1) to  [-32768,32767]

### DIFF
--- a/examples/aishell/s0/conf/train_conformer.yaml
+++ b/examples/aishell/s0/conf/train_conformer.yaml
@@ -41,7 +41,7 @@ raw_wav: true
 collate_conf:
     # waveform level config
     wav_distortion_conf:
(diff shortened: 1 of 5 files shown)


examples/aishell/s0/conf/train_conformer.yaml
```diff
@@ -41,7 +41,7 @@ raw_wav: true
 collate_conf:
     # waveform level config
     wav_distortion_conf:
-        wav_dither: 0 # x means x/32768.0 in torchaudio
+        wav_dither: 1.0
```
robin1001, line 44: change it to 0.0, since our verified best conformer recipe do not use dither now.

robin1001 (approved):

PR #116, opened 2021-01-30
[fix] memory leaks

thehlopster:
Allocated array should be deallocated with `delete []` Deallocating it with `delete` can cause memory leaks.
[More info](https://www.viva64.com/en/examples/v611/)


robin1001 (approved):
good catch

PR #169, opened 2021-02-18
[runtime] support both wave scp and single wave

pengzhendong:


runtime/core/bin/decoder_main.cc
```diff
@@ -36,27 +38,34 @@ int main(int argc, char *argv[]) {
   decode_config.chunk_size = FLAGS_chunk_size;
   wenet::FeaturePipelineConfig feature_config;
   feature_config.num_bins = FLAGS_num_bins;
+  const int sample_rate = 16000;
+  auto feature_pipeline =
+      std::make_shared<wenet::FeaturePipeline>(feature_config);
+
```
robin1001, line 44: Add check  here to ensure wav_path or wav_scp are not empty at the same time

runtime/core/bin/decoder_main.cc
```diff
@@ -3,6 +3,7 @@
 
 #include <chrono>
 #include <iomanip>
+#include <unordered_map>
```
robin1001, line 6: use vector to keep the original order

robin1001 (changes requested):
please see inline

robin1001 (approved):

PR #182, opened 2021-02-24
[recipe] support multi_cn recipe

fanlu:
support multi_cn recipe
aidatatang_url=www.openslr.org/resources/62
aishell_url=www.openslr.org/resources/33
magicdata_url=www.openslr.org/resources/68
primewords_url=www.openslr.org/resources/47
stcmds_url=www.openslr.org/resources/38
thchs_url=www.openslr.org/resources/18


examples/multi_cn/s0/run.sh
```diff
@@ -0,0 +1,237 @@
...
+        cp data/${x}/text data/${x}/text.org
+        paste -d " " <(cut -f 1 -d" " data/${x}/text.org) <(cut -f 2- -d" " data/${x}/text.org | tr -d " ") \
+            > data/${x}/text
+        rm data/${x}/text.org
+        cp -r data/$x $feat_dir
+    done
+    
+    for x in ${test_sets}; do
+        cp data/${x}/test/text data/${x}/test/text.org
+        paste -d " " <(cut -f 1 -d" " data/${x}/test/text.org) <(cut -f 2- -d" " data/${x}/test/text.org | tr -d " ") \
```
robin1001, line 100: any special process of english words here?
I think it is required here, for example, "how are you" should be processed to "how▁are▁you", where ▁ is a special char for space between english words.

examples/multi_cn/s0/local/magicdata_data_filter.py
```diff
@@ -0,0 +1,40 @@
+#!/usr/bin/env python3
```
robin1001, line 1: why not just use filter_scp.pl

robin1001 (changes requested):
please see inline

examples/multi_cn/s0/local/word_segment.py
```diff
@@ -0,0 +1,15 @@
+#!/usr/bin/env python2.7
+#coding:utf-8
```
robin1001, line 2: what's the difference to word_segementation.py?

examples/multi_cn/s0/run.sh
fanlu: Maybe you are right.
But in kaldi's recipe, they transfer all Chinese words and English words to CMU phoneme, But do not add special char for space in words of English, If we add special char, How about read two or more words without interval? like "Not at all".  and There are a lot of words like 'NBA' was read as 'N' 'B' 'A'. I am not sure what is useful for these cases in End2End training.

examples/multi_cn/s0/local/magicdata_data_filter.py
fanlu: Just copy the data preparing script from kaldi's egs/multi_cn

examples/multi_cn/s0/local/word_segment.py
fanlu: word_segment.py use mmseg with python2.7, word_segmentation.py use jieba with python2 or python3

examples/multi_cn/s0/run.sh
robin1001: Really great job.  E2E is different to kaldi where space doesn't require special process. If we don't special symbol for space, we can't know the english word boundary in inference since we doesn't train it at all.

examples/multi_cn/s0/local/word_segment.py
robin1001: I see, word segment is not required for e2e, better to remove it?

examples/multi_cn/s0/run.sh
fanlu: Maybe I can use spm_encode from librispeech for the English part?

examples/multi_cn/s0/local/word_segment.py
fanlu: Got it

fanlu:
> please see inline

fixed

robin1001:
Thank you very much, @whiteshirt0429 please take this PR.

PR #187, opened 2021-02-27
[fix] add explicit utf-8 encoding for dict and output file

chwma0:
[fix] add explicit utf-8 encoding for dict and output file


chwma0:
add more for this reason

PR #197, opened 2021-03-02
[runtime] support arpa lm

robin1001:


robin1001:
> super cool! question: have you evaluated this internally yet? do you see a clear and consistent improvement in the online case?

not yet, it's still on developing. we will release the result once it is ready.

widdiot:
Can this be used for contextual biasing?

robin1001:
> Can this be used for contextual biasing?

No, it's different to contextual biasing.

PR #211, opened 2021-03-15
[feat] add post process for enu because of using bpe model

chwma0:


robin1001 (changes requested):
please see inline

runtime/core/decoder/torch_asr_decoder.cc
```diff
@@ -214,6 +214,39 @@ void TorchAsrDecoder::AttentionRescoring() {
       result_ = result;
     }
   }
+
+  if (opts_.bpe_decode && result_.size()) {
```
robin1001, line 218: Make it a function?

runtime/core/decoder/torch_asr_decoder.cc
```diff
@@ -214,6 +214,39 @@ void TorchAsrDecoder::AttentionRescoring() {
...
+    using std::string;
+    using std::vector;
+    vector<string> characters;
+
+    if (SplitUTF8String(result_, &characters)) {
+      string result;
+      const string delim = "\xe2\x96\x81";
```
robin1001, line 225: Make it a global constant variable in utils.h, kBlankSymbol?

chwma0: OK

runtime/core/decoder/torch_asr_decoder.h
```diff
@@ -46,6 +47,7 @@ class TorchAsrDecoder {
   // Return true if we reach the end of the feature pipeline
   bool AdvanceDecoding();
   void AttentionRescoring();
+  void PostProcesser();
```
robin1001, line 50: change to ProcessBlank()?  and add some comment and example here, such as
// Replace special blank  symbol "▁" (U+2581, \xe2\x96\x81) with white space.
PostProcesser is more than just just blank processing.

runtime/core/decoder/torch_asr_decoder.h
```diff
@@ -24,6 +24,7 @@ using TorchModule = torch::jit::script::Module;
 struct DecodeOptions {
   int chunk_size = 16;
   int num_left_chunks = -1;
+  bool bpe_decode = true;
```
robin1001, line 27: bpe_decode is not required here, we also use ▁ for white space in character based training.

robin1001 (changes requested):
please see inline.

robin1001 (approved):

PR #214, opened 2021-03-15
step start from 1

zh794390558:


wenet/utils/executor.py
```diff
@@ -26,7 +26,7 @@ def train(self, model, optimizer, scheduler, data_loader, device, writer,
                      'larger than before'.format(accum_grad))
         num_seen_utts = 0
         num_total_batch = len(data_loader)
-        for batch_idx, batch in enumerate(data_loader):
+        for batch_idx, batch in enumerate(data_loader, start=1):
```
robin1001, line 29: why start from 1 here?

PR #218, opened 2021-03-17
[network] add bidirectional rescore

whiteshirt0429:


wenet/transformer/decoder.py
```diff
@@ -169,3 +178,153 @@ def forward_one_step(
         if self.use_output_layer:
             y = torch.log_softmax(self.output_layer(y), dim=-1)
         return y, new_cache
+
+class BidirectionalDecoder(TransformerDecoder):
```
robin1001, line 182: rename to BiTransformerDecoder in case that we have other type of decoder in the future.

wenet/transformer/asr_model.py
```diff
@@ -109,31 +122,45 @@ def forward(
         else:
             loss = self.ctc_weight * loss_ctc + (1 -
                                                  self.ctc_weight) * loss_att
-        return loss, loss_att, loss_ctc
+        return loss, loss_att, l_loss_att, r_loss_att, loss_ctc
```
robin1001, line 125: There is no need to return this if it is just used for debug.

robin1001: otherwise there will be more and more return parameters in the future.

wenet/transformer/decoder.py
```diff
@@ -61,13 +63,14 @@ def __init__(
...
-            raise ValueError(
-                f"only 'embed' is supported: {input_layer}")
+            raise ValueError(f"only 'embed' is supported: {input_layer}")
 
         self.normalize_before = normalize_before
         self.after_norm = torch.nn.LayerNorm(attention_dim, eps=1e-12)
         self.use_output_layer = use_output_layer
         self.output_layer = torch.nn.Linear(attention_dim, vocab_size)
+        self.num_blocks = num_blocks
+        self.r_num_blocks = 0
```
robin1001, line 73: remove this

wenet/transformer/asr_model.py
```diff
@@ -622,6 +686,10 @@ def init_asr_model(configs):
     vocab_size = configs['output_dim']
 
     encoder_type = configs.get('encoder', 'conformer')
+    decoder_type = configs.get('decoder', 'transformer')
+    assert encoder_type in ["transformer", "conformer"]
+    assert decoder_type in ["transformer", "bidirectional"]
```
robin1001, line 691: bitransformer

wenet/utils/common.py
```diff
@@ -84,6 +111,52 @@ def add_sos_eos(ys_pad: torch.Tensor, sos: int, eos: int,
     return pad_list(ys_in, eos), pad_list(ys_out, ignore_id)
 
 
+def add_eos_sos(ys_pad: torch.Tensor, sos: int, eos: int,
```
robin1001, line 114: eos and sos has the same meaning in both forward and backward decoder, there is no need to add the new function.

wenet/transformer/decoder.py
```diff
@@ -90,29 +91,55 @@ def forward(
         memory_mask: torch.Tensor,
         ys_in_pad: torch.Tensor,
         ys_in_lens: torch.Tensor,
-    ) -> Tuple[torch.Tensor, torch.Tensor]:
+        r_ys_in_pad: Optional[torch.Tensor] = None,
+        reverse: bool = False,
```
robin1001, line 95: make reverse a member of transformer

wenet/transformer/decoder.py
```diff
@@ -90,29 +91,55 @@ def forward(
...
+            assert r_tgt is not None
+            # r_tgt_mask: (B, 1, L)
+            tgt_mask = (~make_pad_mask_right(ys_in_lens).unsqueeze(1)).to(
+                tgt.device)
+            # r_m: (1, L, L)
+            m = subsequent_mask_right_to_left(
+                tgt_mask.size(-1), device=tgt_mask.device).unsqueeze(0)
+
+            # r_tgt_mask: (B, L, L)
+            tgt_mask = tgt_mask & m
```
robin1001, line 142: common code with above if section

PR #231, opened 2021-03-24
[runtime] fix bug and print timestamp

pengzhendong:


runtime/core/decoder/torch_asr_decoder.cc
```diff
@@ -218,33 +212,49 @@ void TorchAsrDecoder::AttentionRescoring() {
...
 
       // Ignore tailing space
-      if (result.size() && result.back() != ' ') {
+      if (!result.empty() && result.back() == ' ') {
         result.pop_back();
       }
 
-      for (size_t i = 0; i < result.size(); ++i) {
-        result[i] = tolower(result[i]);
+      for (char& ch : result) {
```
robin1001, line 256: it is not preferred to change ch in this way.

runtime/core/decoder/torch_asr_decoder.cc
```diff
@@ -218,33 +212,49 @@ void TorchAsrDecoder::AttentionRescoring() {
...
+    std::vector<std::string> characters;
 
     if (SplitUTF8String(result_, &characters)) {
-      string result;
-
-      for (size_t i = 0; i < characters.size(); ++i) {
-        if (characters[i] != wenet::kSpaceSymbol) {
-          result.append(characters[i]);
+      std::string result;
+      for (auto& ch : characters) {
```
robin1001, line 240: use auto as less as possible

runtime/core/bin/decoder_main.cc
```diff
@@ -110,9 +110,12 @@ int main(int argc, char *argv[]) {
       }
     }
     LOG(INFO) << "Final result: " << decoder.result();
+    LOG(INFO) << "Final timestamp in ms: " << decoder.timestamp();
     LOG(INFO) << "Decoded " << wave_dur << "ms audio taken " << decode_time
               << "ms.";
-    buffer << wav.first << " " << decoder.result() << std::endl;
+    buffer << wav.first << "\t"
```
robin1001, line 116: why output timestamp here?

PR #233, opened 2021-03-25
[fix] fix some errors perhaps happen in large dataset training

whiteshirt0429:


wenet/utils/executor.py
```diff
@@ -81,7 +81,8 @@ def cv(self, model, data_loader, device, args):
         '''
         model.eval()
         log_interval = args.get('log_interval', 10)
-        num_seen_utts = 0
+        # in order to the denominator being 0
+        num_seen_utts = 0.0001
```
robin1001, line 85: num_seen_utts = 1

robin1001, line 85: in order to avoid division by 0

robin1001 (approved):

PR #238, opened 2021-03-25
[tools] add ctc alignment

whiteshirt0429:


wenet/transformer/ctc.py
```diff
@@ -67,3 +68,63 @@ def argmax(self, hs_pad: torch.Tensor) -> torch.Tensor:
             torch.Tensor: argmax applied 2d tensor (B, Tmax)
         """
         return torch.argmax(self.ctc_lo(hs_pad), dim=2)
+
+
+    def forced_align(self, ctc_probs: torch.Tensor, y: torch.Tensor,
+                     blank_id=0) -> list:
```
robin1001, line 74: why add this function in CTC module, it can be a standalone function since it is not related to CTC module.

wenet/transformer/ctc.py
```diff
@@ -67,3 +68,63 @@ def argmax(self, hs_pad: torch.Tensor) -> torch.Tensor:
...
+            torch.Tensor ctc_probs: hidden state sequence, 2d tensor (T, D)
+            torch.Tensor y: id sequence tensor 1d tensor (L)
+            int blank_id: blank symbol index
+        Returns:
+            torch.Tensor: alignment result
+        """
+
+        lpz = ctc_probs
+
+        y_int = insert_blank(y, blank_id)
```
robin1001, line 87: why named y_int here?

tools/alignment.sh
```diff
@@ -0,0 +1,49 @@
+#!/bin/bash
```
robin1001, line 1: Is it better to support alignment in the end of run.sh on test set? since different dataset may have different format data pipeline.

wenet/transformer/ctc.py
```diff
@@ -67,3 +69,67 @@ def argmax(self, hs_pad: torch.Tensor) -> torch.Tensor:
...
+
+        Args:
+            torch.Tensor ctc_probs: hidden state sequence, 2d tensor (T, D)
+            torch.Tensor y: id sequence tensor 1d tensor (L)
+            int blank_id: blank symbol index
+        Returns:
+            torch.Tensor: alignment result
+        """
+
+        ctc_hyps = ctc_probs
```
robin1001, line 87: just use ctc_probs is okay. ctc_hyps is ambiguous, typically, hypotheses is for recognition result.

wenet/transformer/ctc.py
```diff
@@ -67,3 +69,67 @@ def argmax(self, hs_pad: torch.Tensor) -> torch.Tensor:
...
+
+        y_insert_blank = insert_blank(y, blank_id)
```
robin1001, line 89: maybe ys with comment is okay instead of y_insert_blank, which will make small code

robin1001 (approved):

PR #242, opened 2021-03-27
[runtime] refactor prefix score

pengzhendong:


runtime/core/decoder/ctc_prefix_beam_search.h
```diff
@@ -31,7 +31,11 @@ struct PrefixScore {
   float cur_token_prob = -kFloatMax;  // prob of current token
   std::vector<int> times_s;           // times of viterbi blank path
   std::vector<int> times_ns;          // times of viterbi none blank path
+
   PrefixScore() = default;
+  float score() const { return LogAdd(s, ns); }
+  float viterbi_score() const { return v_s > v_ns ? v_s : v_ns; }
+  std::vector<int> times() const { return v_s > v_ns ? times_s : times_ns; }
```
robin1001, line 38: const std::vector<int>& times()

runtime/core/decoder/ctc_prefix_beam_search.cc
```diff
@@ -42,9 +42,7 @@ static bool PrefixScoreCompare(
 }
```
robin1001, line 42: Great simplification here! Great job

runtime/core/decoder/ctc_prefix_beam_search.h
pengzhendong: fixed.

robin1001 (approved):

PR #243, opened 2021-03-28
Update for training on GPU 3090

jinsongpan:
## what's the purpose of the changes
 Unable to train normally on GPU3090 using the latest commit.

## modified:
tip1: The default backends of torchaudio>=0.8.0 is changed to"sox_io"; so need set the audio backend as "sox" for normal training; 
ref: https://pytorch.org/audio/stable/backend.html#backend  for full details.

tips2: "torchaudio.sox_effects.SoxEffectsChain()" is invalid in torchaudio>=0.8.0

tip3: yaml.load() without Loader=... is deprecated in yaml>5.1,  as the default Loader is unsafe;
ref: https://msg.pyyaml.org/load for full details.

### DIFF
--- a/examples/multi_cn/s0/run.sh
+++ b/examples/multi_cn/s0/run.sh
@@ -93,8 +93,6 @@ if [ ${stage} -le 0 ] && [ ${stop_stage} -ge 0 ]; then
     local/stcmds_data_prep.sh $dbase/stcmds data/stcmds || exit 1;
     local/tal_data_prep.sh $dbase/TAL/TAL_ASR data/tal_asr || exit 1;
     local/tal_mix_data_prep.sh $dbase/TAL/TAL_ASR_mix data/tal_mix || exit 1;
-    local/tal_data_prep.sh $dbase/tal/tal_asr_data data/tal_asr || exit 1;
-    local/tal_mix_data_prep.sh $dbase/tal/tal_asr_mix_data data/tal_mix || exit 1;
 
     if $has_aishell2; then
         local/aishell2_data_prep.sh $dbase/aishell2/IOS data/aishell2/train || exit 1;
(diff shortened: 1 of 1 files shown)


robin1001:
The updated API may works for torchaudio>=0.80, but it should fail for lower version. Maybe we can just specify the torchaudio with lower version

robin1001:
I specified the torchaudio version in https://github.com/mobvoi/wenet/pull/245, and your changes may be adopted in the future for a big version update.

jinsongpan:
> I specified the torchaudio version in #245, and your changes may be adopted in the future for a big version update.

Sorry, i didn't notice that "torchaudio.sox_effects.apply_effects_file()" did not supperted on torchaudio==0.6.0;
but i don't think  #245 is a good idea, because it still don't support GPU 3090; I advice not fixing the version to torchaudio==0.6.0 and adding a few lines of code in the dataset.py: if torchaudio<=0.6.0, using "SoxEffectsChain()"; else using "apply_effects_file()".
other changes are useful in lower version.

robin1001:
Do you mean torchaudio 0.60 is not supported on 3090? Or torchaudio 0.60 could not cowork with higher pytorch version1.8?

wenet/dataset/dataset.py
```diff
@@ -1,4 +1,5 @@
 # Copyright (c) 2020 Mobvoi Inc. (authors: Binbin Zhang, Chao Yang)
+# 2021 jinsong pan
```
robin1001, line 2: Copyright (c) 2021 Jinsong Pan

jinsongpan:
> Do you mean torchaudio 0.60 is not supported on 3090? Or torchaudio 0.60 could not cowork with higher pytorch version1.8?

yes. I have updated the code to suport all torchaudio versions until now.

PR #261, opened 2021-03-30
fix bugs to support torchaudio==0.8.0

jinsongpan:
ref: [#243](https://github.com/mobvoi/wenet/pull/243)


jinsongpan:
deprecated

PR #265, opened 2021-03-30
[runtime] support endpoint

pengzhendong:


runtime/core/decoder/torch_asr_decoder.cc
```diff
@@ -120,8 +124,8 @@ bool TorchAsrDecoder::AdvanceDecoding() {
                                       ->run_method("ctc_activation", chunk_out)
                                       .toTensor()[0];
     encoder_outs_.push_back(std::move(chunk_out));
-    ctc_prefix_beam_searcher_->Search(ctc_log_probs);
-    UpdateResult();
+    UpdateResult(ctc_log_probs);
+    finish = IsEndpoint(ctc_log_probs) || finish;
```
robin1001, line 129: It's better if we can create a class named CtcEndpointer, which has it's own state related to endpoint, and accept ctc_log_probs as input, and it has the function like EndpointDetected(int frame), just like what we do in recognizer.

runtime/core/decoder/torch_asr_decoder.cc
```diff
@@ -120,8 +124,8 @@ bool TorchAsrDecoder::AdvanceDecoding() {
...
```
robin1001, line 128: We should add more states tag to show what state we are.
1. normal decoding state
2. end of feature
3. endpoint

runtime/core/websocket/websocket_server.h
```diff
@@ -60,14 +59,15 @@ class ConnectionHandler {
...
-  int nbest_;
+  int nbest_ = 1;
   websocket::stream<tcp::socket> ws_;
   std::shared_ptr<FeaturePipelineConfig> feature_config_;
   std::shared_ptr<DecodeOptions> decode_config_;
   std::shared_ptr<SymbolTable> symbol_table_;
   std::shared_ptr<TorchAsrModel> model_;
 
   bool got_start_tag_ = false;
+  bool got_end_tag_ = false;
```
robin1001, line 70: change it to stop_recognition_
and add some comment such as
// When endpoint is detected, we should stop recognition and stop receiving data.

robin1001 (approved):

PR #266, opened 2021-03-30
fix bugs to support torchaudio==0.8.0

jinsongpan:
ref: [#243](https://github.com/mobvoi/wenet/pull/243)

### DIFF
--- a/wenet/dataset/dataset.py
+++ b/wenet/dataset/dataset.py
@@ -1,4 +1,5 @@
 # Copyright (c) 2020 Mobvoi Inc. (authors: Binbin Zhang, Chao Yang)
+# Copyright (c) 2021 Jinsong Pan
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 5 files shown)


wenet/dataset/dataset.py
```diff
@@ -144,15 +148,25 @@ def _load_wav_with_speed(wav_file, speed):
         return torchaudio.load_wav(wav_file)
     else:
         si, _ = torchaudio.info(wav_file)
-        E = torchaudio.sox_effects.SoxEffectsChain()
-        E.append_effect_to_chain('speed', speed)
-        E.append_effect_to_chain("rate", si.rate)
-        E.set_input_file(wav_file)
-        wav, sr = E.sox_build_flow_effects()
+
+        if int(torchaudio.__version__.split(".")[1]) < 8:
```
robin1001, line 152: better method for it？

jinsongpan: have fixed.

PR #275, opened 2021-04-02
torch rescore empty tensor bug fix

Mddct:
when hyp num is zero , it will cause core dump , see below:
截屏2021-04-02 下午3 16 50


robin1001:
please refine the lint problem.

robin1001 (approved):

PR #300, opened 2021-04-09
Update fbank.conf

double22a:
Consider some cases where the sample rate is not 16k


robin1001:
it is not required for librispeech.

PR #306, opened 2021-04-12
support subword prefix specific

Mddct:
user can specific subword prefix such as "##" or "_"


robin1001:
What symbol do you use for space?

Mddct:
I use ”##”。 Because spm can denote a space indicator, and bert tokensize also use ##。

For example：

他非常friendly -> 他 非 常 ##fri end ly

robin1001:
Please note this logic suppose only one UTF8 character "▁" (U+2581) is used to denote <space>,  If there are more than one character, it may gives wrong result.

for example, ## is used for space,  then both ## and # will be replaced to blank.

I think maybe you can just replace ## with "▁ " in your symbol table file.

Mddct:
It's an easy way. then I close this pr. Thanks

PR #310, opened 2021-04-15
[fix] subsample mask fix. it does not effects models preformance if y…

whiteshirt0429:
…ou sort by audio length during training process


whiteshirt0429:
here it is a draft version, since it indeed does not influence model performance when sorting the audio seqs.

PR #314, opened 2021-04-19
[runtime] surport tlg lm

robin1001:


mikelei:
Better to add a unit test for decoding.

robin1001:
> Better to add a unit test for decoding.

okay, we will try to add it later. 
it's a draft PR and it's still in early stage now. it will be separated into several small PRs once it is ready.

Mddct:
If we regard blank as sil, then the timestamp can be obtained by decoding in wfst, so there is no need to call ctcendpoint afterwards?

robin1001:
> If we regard blank as sil, then the timestamp can be obtained by decoding in wfst, so there is no need to call ctcendpoint afterwards?

It's more flexible to keep decoder and endpoint independent.

PR #321, opened 2021-04-23
[runtime] add search interface

robin1001:


pengzhendong (approved):

runtime/core/decoder/torch_asr_decoder.cc
```diff
@@ -143,7 +150,18 @@ DecodeState TorchAsrDecoder::AdvanceDecoding() {
                                       ->run_method("ctc_activation", chunk_out)
                                       .toTensor()[0];
     encoder_outs_.push_back(std::move(chunk_out));
-    UpdateResult(ctc_log_probs);
+    auto search_start = std::chrono::steady_clock::now();
```
XiangmingWang, line 153: why not consider about Kaldi::Timer?

XiangmingWang (approved):

runtime/core/decoder/torch_asr_decoder.cc
robin1001: Will refine it later

PR #324, opened 2021-04-25
[runtime] refine flags

robin1001:


pengzhendong (approved):

runtime/core/decoder/params.h
```diff
@@ -0,0 +1,63 @@
...
+std::shared_ptr<DecodeOptions> InitDecodeOptionsFromFlags() {
+  auto decode_config = std::make_shared<DecodeOptions>();
+  decode_config->chunk_size = FLAGS_chunk_size;
+  decode_config->num_left_chunks = FLAGS_num_left_chunks;
+  decode_config->ctc_weight = FLAGS_ctc_weight;
+  decode_config->rescoring_weight = FLAGS_rescoring_weight;
+  return decode_config;
+}
+
+std::shared_ptr<fst::SymbolTable> InitSymbolTableFromFlags() {
```
XiangmingWang, line 55: move symbol table  into model would be better?

XiangmingWang (changes requested):

runtime/core/decoder/params.h
robin1001: what model?

XiangmingWang (approved):

PR #333, opened 2021-04-25
[runtime] add wfst decoder

robin1001:


runtime/core/decoder/ctc_wfst_beam_search.cc
```diff
@@ -0,0 +1,141 @@
+// Copyright 2020 Mobvoi Inc. All Rights Reserved.
```
XiangmingWang, line 1: copyright 2021

runtime/core/decoder/ctc_wfst_beam_search.h
```diff
@@ -0,0 +1,91 @@
...
+  int32 NumIndices() const override;
+  void AcceptLoglikes(const torch::Tensor& logp);
+  void SetFinish() { done_ = true; }
+
+ private:
+  int num_frames_ready_ = 0;
+  float scale_ = 1.0;
+  bool done_ = false;
+  torch::Tensor logp_;
+  std::unique_ptr<torch::TensorAccessor<float, 1>> accessor_ = nullptr;
```
XiangmingWang, line 38: smart ptr will be nullptr by default.

runtime/core/decoder/params.h
```diff
@@ -43,12 +52,26 @@ std::shared_ptr<FeaturePipelineConfig> InitFeaturePipelineConfigFromFlags() {
...
   auto decode_config = std::make_shared<DecodeOptions>();
   decode_config->chunk_size = FLAGS_chunk_size;
   decode_config->num_left_chunks = FLAGS_num_left_chunks;
   decode_config->ctc_weight = FLAGS_ctc_weight;
   decode_config->rescoring_weight = FLAGS_rescoring_weight;
+  decode_config->ctc_wfst_search_opts.max_active = FLAGS_max_active;
+  decode_config->ctc_wfst_search_opts.beam = FLAGS_beam;
+  decode_config->ctc_wfst_search_opts.lattice_beam = FLAGS_lattice_beam;
+  decode_config->ctc_wfst_search_opts.acoustic_scale = FLAGS_acoustic_scale;
+  decode_config->ctc_wfst_search_opts.blank_skip_thresh =
```
XiangmingWang, line 73: consider to use proto instead

runtime/core/decoder/torch_asr_decoder.cc
```diff
@@ -220,6 +219,9 @@ void TorchAsrDecoder::UpdateResult() {
 void TorchAsrDecoder::AttentionRescoring() {
   searcher_->FinalizeSearch();
   UpdateResult();
+  if (0.0 == opts_.rescoring_weight) {
```
XiangmingWang, line 222: we'd better not compare two float-pointing value, fabs(a - b) < 1e-6 would be safer

XiangmingWang (changes requested):

AdolfVonKleist:
question out of curiosity: are you moving towards support for features like contextual biasing, or just aiming to apply the wfst decoders for complementary LM support?

robin1001:
> question out of curiosity: are you moving towards support for features like contextual biasing, or just aiming to apply the wfst decoders for complementary LM support?

WFST based LM support is our first step, and contextual biasing is in our further plan.

runtime/core/decoder/torch_asr_decoder.cc
robin1001: Here we means exact zero, not close to 0. Actually we write in this way since we see a lot of code like this in kaldi and other open source c++ project.

runtime/core/decoder/params.h
robin1001: Maybe add it in the future.

runtime/core/decoder/ctc_wfst_beam_search.h
robin1001: fixed

runtime/core/decoder/ctc_wfst_beam_search.cc
robin1001: fixed

XiangmingWang (approved):

pengzhendong (approved):

AdolfVonKleist:
> WFST based LM support is our first step, and contextual biasing is in our further plan.
It looked like you also are including lexicon-related scripts.  Is that just incidental, or do you also plan to require pronunciation lexica in future releases in order to use these features?

robin1001:
Yes, we are going to require pronunciation lexicon if LM is required. However, the lexicon is different to what is used to be, it's a character based or BPE based lexicon, just for convenient to build the decoding graph with LM. For char based model, the lexicon is like:

apple     _ a p p l e
中国      中  国

PR #335, opened 2021-04-26
flush cmvn log in time

dophist:
1. change logging from stdout to stderr
2. Training pipeline seems halted without instant logging flush, especially when data-set is large.

### DIFF
--- a/tools/compute_cmvn_stats.py
+++ b/tools/compute_cmvn_stats.py
@@ -1,6 +1,7 @@
 #!/usr/bin/env python3
 # encoding: utf-8
 
+import sys
 import argparse
 import json
 import codecs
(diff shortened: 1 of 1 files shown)


robin1001 (approved):

tools/compute_cmvn_stats.py
```diff
@@ -90,7 +91,7 @@ def __getitem__(self, idx):
             all_number += number
             wav_number += batch_size
             if wav_number % 1000 == 0:
-                print('process {} wavs,{} frames'.format(wav_number, all_number))
+                print(f'processed {wav_number} wavs, {all_number} frames', file=sys.stderr, flush=True)
```
robin1001, line 94: the line more than 80 character and the lint failed

robin1001 (changes requested):
please see inline

robin1001 (approved):

PR #338, opened 2021-04-26
k2 style wfst_based_decoding

xu-gaopeng:
This PR add WFST-based decoding to E2E models

1.add functions for dealing with WFST
2.integrate K2 create token/lexicon/grammar fst, composing decoding graph
3.add scrips to make graph (to do)

￼

### DIFF
--- a/wenet/bin/recognize.py
+++ b/wenet/bin/recognize.py
@@ -27,6 +27,7 @@ from torch.utils.data import DataLoader
 from wenet.dataset.dataset import AudioDataset, CollateFunc
 from wenet.transformer.asr_model import init_asr_model
 from wenet.utils.checkpoint import load_checkpoint
+import k2
 
 if __name__ == '__main__':
     parser = argparse.ArgumentParser(description='recognize with your model')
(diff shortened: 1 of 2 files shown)


wenet/bin/recognize.py
```diff
@@ -126,6 +129,19 @@
             assert len(arr) == 2
             char_dict[int(arr[1])] = arr[0]
     eos = len(char_dict) - 1
+    #Load HLG  for wfst-based decoding
+    if args.mode == 'wfst_based_decoding':
+        d = torch.load('data/graph/HLG.pt')
```
fanlu, line 134: it's better to add how HLG was built.

robin1001:
Thanks for your contributing. It's great to see wenet works with k2. It's a great example and reference if someone wants to do the same works on k2 like this.

However, currently we have no plan to support k2 style FST in wenet. We are integrating Kaldi style WFST directly to our c++ runtime to support LM. please see:
https://github.com/mobvoi/wenet/pull/339
https://github.com/mobvoi/wenet/pull/336

So we are not going to merge it, and let's keep it as a PR for k2 reference.

wenet/bin/recognize.py
xu-gaopeng: Yes, I will finish it in the near future

PR #347, opened 2021-04-29
Update train.py

double22a:
print the number of model parameter

### DIFF
--- a/wenet/bin/train.py
+++ b/wenet/bin/train.py
@@ -164,7 +164,7 @@ if __name__ == '__main__':
     # Init asr model from configs
     model = init_asr_model(configs)
     print(model)
-    print('num. model params: %s' % sum(p.numel() for p in model.parameters()))
+    print('the number of model params: %s' % sum(p.numel() for p in model.parameters()))
 
     # !!!IMPORTANT!!!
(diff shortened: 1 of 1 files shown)


wenet/bin/train.py
```diff
@@ -164,6 +164,7 @@
     # Init asr model from configs
     model = init_asr_model(configs)
     print(model)
+    print('num. model params: %s' % sum(p.numel() for p in model.parameters()))
```
robin1001, line 168: fix num. to num

wenet/bin/train.py
```diff
@@ -164,6 +164,7 @@
     # Init asr model from configs
     model = init_asr_model(configs)
     print(model)
+    print('the number of model params: %s' % sum(p.numel() for p in model.parameters()))
```
robin1001, line 167: print('the number of model params: {}'.format(sum(p.numel() for p in model.prameters())), string with format is preferred in wenet

robin1001 (changes requested):
please see inline

robin1001:
the line is too long, more than 80 character, please fix it.

robin1001 (approved):

PR #355, opened 2021-05-03
[tools] add segmented wav.scp support

whiteshirt0429:


wenet/dataset/dataset.py
```diff
@@ -199,10 +201,26 @@ def _extract_feature(batch, speed_perturb, wav_distortion_conf,
...
         try:
+            wav = x[1]
+            value = wav.strip().split(":")
+            assert len(value) >= 1
+            wav_path = value[0]
+            sample_rate = torchaudio.backend.sox_backend.info(wav_path)[0].rate
             if speed_perturb:
-                waveform, sample_rate = _load_wav_with_speed(x[1], speed)
+                assert len(value) == 1
+                waveform, sample_rate = _load_wav_with_speed(wav_path, speed)
```
robin1001, line 211: segments should also be supported in speed_perturb

tools/format_data.sh
```diff
@@ -76,9 +77,14 @@ if [ -n "${feat}" ]; then
                 --filetype "${filetype}" \
                 --preprocess-conf "${preprocess_conf}" \
                 --verbose ${verbose} ${feat} ${tmpdir}/input_${i}/shape.scp
-        elif [ ${feat_type} == "wav" ] || [ ${feat_type} == "flac" ]; then
-            tools/wav_to_duration.sh --nj ${nj} \
-                ${feat} ${tmpdir}/input_${i}/shape.scp
+        elif [ ${feat_type} == "wav" ] || [ ${feat_type} == "flac" ] || [ ${feat_type} == "opus" ]; then
+            if [ ! -f $dir/utt2dur ]; then
```
hehehhh, line 81: awk '{print $1" "$4-$3}' segments > utt2dur

PR #368, opened 2021-05-10
[doc] update s0 librispeech  result

whiteshirt0429:


robin1001 (changes requested):
please realign the table separator

robin1001 (approved):

PR #378, opened 2021-05-12
[Add] Support swap the order of the convolution module and the attention.

TeaPoly:
Support swapping the order of the convolution module and the multi-head self-attention module, as described in https://arxiv.org/pdf/2011.10798.pdf.


TeaPoly:
The diagram from paper is here:
截屏2021-05-12 下午12 25 46

TeaPoly:
Need for experiments to confirm.

PR #394, opened 2021-05-19
add grpc interface

aluminumbox:
Many developers comment in wenet wechat group, consulting about industrialized production-ready service.
Socket is indeed the most common way to call streaming ASR service. And it is used in many company like Iflytek.
However, if one considers the Microservice architecture, we may need an industrialized RPC architecture.
About why using grpc, you may refer to https://grpc.io/ 
It supports streaming service based on HTTP2.0. It supports concurrency control and queue.
It is fast, and supports multiple language like python/java/c++
Here, we contribute the wenet core code with grpc interface.
For example, you can offer socket interface to your customers, and use grpc interface within your company, which is our current service architecture in Ximalaya.
We are glad if our code may be helpful to other folks.


runtime/core/grpc/wenet.proto
```diff
@@ -0,0 +1,20 @@
...
+option java_package = "ex.grpc";
+option objc_class_prefix = "wenet";
+
+package wenet;
+
+service ASR{
+  rpc Recognize(stream Request) returns (stream Response) {}
+}
+
+message Request{
```
robin1001, line 12: add space here

runtime/core/grpc/wenet.proto
```diff
@@ -0,0 +1,20 @@
...
+  int32 nbest = 1;
+  bool continuous_decoding = 2;
+  bytes audio_data = 3;
+}
+
+message Response{
+  string response_json = 1;
```
robin1001, line 19: if we use grpc, we should specify all the fields directly here instead of wrap it in the json string, right?

robin1001 (commented):
please see inline.

runtime/core/grpc/wenet.proto
aluminumbox: We used a more complex proto, which gives explicit definition to all fields now. However, it may cause some inconvenience to folks who are new to protobuf.

runtime/core/bin/grpc_client_main.cc
```diff
@@ -0,0 +1,65 @@
+// Copyright (c) 2020 Ximalaya Inc (Xiang Lyu)
```
robin1001, line 1: 2021

runtime/server/x86/CMakeLists.txt
```diff
@@ -19,6 +20,16 @@ if(CMAKE_CXX_COMPILER_ID MATCHES "GNU|Clang")
   set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++14 -pthread -fPIC")
 endif()
 
+# third_party: grpc
+# On how to build grpc, you may refer to https://github.com/grpc/grpc
+# We recommend manually recursive clone the repo to avoid internet connection problem
+execute_process(COMMAND apt-get install -y build-essential autoconf libtool pkg-config)
```
robin1001, line 26: remove this, otherwise it will fails when we don't have the sudo access.

robin1001 (approved):

PR #414, opened 2021-05-27
a simpler realization of rel shift

boji123:
a simpler realization of rel shift


robin1001 (commented):
please note rel_shift is not used in our implementation, it has been commented. see the code below.

PR #440, opened 2021-06-06
接受linux shell管道输入（wav二进制） can accept shell pipe input (bytes wav)

Mashiro083:
add process_pipe_input.py into tools/
change wenet/dataset/dataset.py and tools/wav2dur.py

通过subprocess执行shell命令，用scipy.io.wavfile.read来读取二进制输入
利用是否以"|"结尾判断输入为音频还是管道
do shell command by subprocess and get waveform by scipy.io.wavfile.read(io.BytesIO(out))
Use whether the input ends with "|" to determine whether the input is audio or pipe

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -6,3 +6,4 @@ tensorboard
 tensorboardX
 typeguard
 textgrid
+scipy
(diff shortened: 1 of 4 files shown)


robin1001:
If you want to use sox,  just use the sox backend.

PR #447, opened 2021-06-10
[runtime] create bitrev and sintbl for each feature pipeline

GenjunLiu:
fft may crash in multi-threaded scenarios:

e.g.
```
last_n = 0

thread1:
   1) if (n != last_n || n == 0)  pass
   2) make_sintbl/make_bitrev
   3) access bitrev and sintbl

thread2:
   1）if (n != last_n || n == 0)  pass
   2) if (sintbl != NULL) free(sintbl);

```
if the order is:
- thread1: 1
- thread2: 1
- thread1: 2
- thread2: 2
- thread1: 3 (crash because `sintbl` has been freed )

### DIFF
--- a/runtime/core/frontend/fbank.h
+++ b/runtime/core/frontend/fbank.h
@@ -41,6 +41,13 @@ class Fbank {
         distribution_(0, 1.0),
         dither_(0.0) {
     fft_points_ = UpperPowerOfTwo(frame_length_);
+    // generate bit reversal table and trigonometric function table
+    const int fft_points_4 = fft_points_ / 4;
+    bitrev_.resize(fft_points_);
+    sintbl_.resize(fft_points_ + fft_points_4);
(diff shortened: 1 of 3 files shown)


robin1001:
It's caused by static variable, right? The best way to fix it is just allocate the memory on the fly. How do you think?

GenjunLiu:
The simplest way is yes, but it's a frequent operation. I think that's why to use a cache before

robin1001:
Lock is not preferred when we have multiple threads for multiple request, especially in this case, where FFT is called in the whole server/client interaction.

robin1001:
Would you like to refine it in the simplest way, I think it is preferred.

GenjunLiu:
Do you mean the latest version? already removed mutex.

robin1001 (approved):

robin1001:
Good catch and refinement.

PR #464, opened 2021-06-18
[doc] add u2++ lm support and update aishell results

whiteshirt0429:


examples/aishell/s0/run.sh
```diff
@@ -229,9 +231,14 @@ if [ ${stage} -le 7 ] && [ ${stop_stage} -ge 7 ]; then
         data/local/dict data/local/tmp data/local/lang
     tools/fst/make_tlg.sh data/local/lm data/local/lang data/lang_test || exit 1;
     # 7.4 Decoding with runtime
+    # Only U2++ model is used, does reverse_weight can be greater than 0.0.
```
robin1001, line 234: what do you mean?

examples/aishell/s0/run.sh
```diff
@@ -229,9 +231,14 @@ if [ ${stage} -le 7 ] && [ ${stop_stage} -ge 7 ]; then
...
+    # While using U2++ model, only left to right decoder will be used if reverse_weight is 0.0.
```
robin1001, line 235: reverse_weight only works for u2++ model and only left to right decoder is used when it is set to 0.0.

robin1001 (changes requested):
please see inline.

robin1001 (approved):

PR #152, opened 2021-02-07
s0, gpu fbank and resample

qmpzzpmq:

PR #492, opened 2021-07-06
Fbank on GPU

wangtianrui:
I realized FBank calculation of GPU in my project, and the training speed was 28% higher than that of CPU calculation, and the utilization rate of graphics card was also improved

However, due to time constraints, I directly copy the key codes to you. I hope to make my contribution to WENET


robin1001:
There is a same PR for GPU fbank, which is more simple. Please see https://github.com/wenet-e2e/wenet/pull/152

PR #507, opened 2021-07-11
Fix possibly a very small input size limit

ngoel17:


examples/librispeech/s1/run.sh
```diff
@@ -114,7 +114,7 @@ if [ ${stage} -le 3 ] && [ ${stop_stage} -ge 3 ]; then
         # remove utt having more than 400 characters
         tools/remove_longshortdata.py \
             --min_input_len 0.5 \
-            --max_input_len 20 \
+            --max_input_len 3000 \
```
robin1001, line 117: It's second here. Very long data will result in OOM problem in training.

ngoel17: In that case there is some other bug because with these parameters, format.data is empty.

PR #524, opened 2021-07-21
[runtime] pack all resource realted to decoder to DecodeResource to s…

robin1001:
…implify interface design

Change-Id: Id5319b016f8a37fb42832ff105dc0d93294ed729


robin1001:
Here we pack all the static resources related to decoder into DecodeResource struct. which:
1. simplify the interface design, now the interface is more simple and clear.
2. it's easy to extend. There is no need to change the interface if we want to add new resource to decoder. Think about the previous design, we have to modify TorchAsrDecoder, websocket interface, grpc interface if we want to add new resource.

pengzhendong (approved):

PR #525, opened 2021-07-21
[runtime] support timestamp in wfst decoding

robin1001:
Change-Id: Ic53232beb76fa96d2edb5d679ce7ea95f6c876b0


robin1001:
TimeStamp of the output of CtcWfstBeamSearch may be inaccurate due to various FST operations when building the decoding graph. So here we use time stamp of the input(e2e model unit), which is more accurate, and it requires the symbol table of the e2e model used in training.

wenet-libs (approved):

pengzhendong (approved):

PR #618, opened 2021-09-11
[fix(android)] init post_processor in android demo

xingchensong:
fix app crash


xingchensong:
Should we update APK link in README.md ？

robin1001 (approved):

robin1001:
> Should we update APK link in README.md ？

The old APP works fine. Will update it if new features in.

PR #620, opened 2021-09-14
[network] unify io for both small dataset and big dataset 

robin1001:
## Problems

We have to load all the `text` and `audio`(file path) into memory at once, which result in OOM problems for very big data(50k+ hours). And the training is slow due to random audio file access.

## Solutions

Inspired by TFRecord and [webdataset](https://github.com/webdataset/webdataset), we are going to design a new IO to unify small and big data training. The key features of the new IO is:

* Sequence Reading
  *  Shard(tar several small files into a Shard)
  * Local IO and Cloud IO(HDFS/OSS/S3 by HTTP) support
* Random Reading, backward compatibility with the previous random IO.

image

Chain operation is designed since it's:
* extensible
* easy to debug

**And we will remove the old IO once the unified IO shows promising and stable performance.**

## How to use ? 

Please see PR https://github.com/wenet-e2e/wenet/pull/627.

## Questions

1. Why not TFRecord?
> TFRecord depends on a Protobuf, which is heavy, poorly readable, and lacks of ecosystem

2. Why not webdataset?
> webdataset is great, but seems it's not designed for speech task. And we want to support random reading in the old way, which is impossible in webdataset.

3. Why not LMDB, HDF5, Databases? 
> They solved the massive small files problem, but can not solve the random access problem which results in inefficient IO.


jprobichaud:
Have you considered https://github.com/NVIDIA/DALI ?

robin1001:
> Have you considered https://github.com/NVIDIA/DALI ?

I don't know DALL before.  Seems it is too heavy for WeNet.

Mddct:
Can dataset  support map and repeat like tensorflow ?

robin1001:
> Can dataset support map and repeat like tensorflow ?

It's very easy in the new IO. All you need is adding a new processor function.

jprobichaud:
> 
> 
> > Have you considered [NVIDIA/DALI](https://github.com/NVIDIA/DALI) ?
> 
> I don't know DALL before. Seems it is too heavy for WeNet.

I didn't new DALI before either, but they seem to have solve the exact same issue you are trying to. They also support doing some pipeline operations on the GPU and in a distributed fashion.  Perhaps it's worth having a good look.  They did the work for the Jasper ASR, so they are aware of the speech recognition context as well.

robin1001:
> > > Have you considered [NVIDIA/DALI](https://github.com/NVIDIA/DALI) ?
> > 
> > 
> > I don't know DALL before. Seems it is too heavy for WeNet.
> 
> I didn't new DALI before either, but they seem to have solve the exact same issue you are trying to. They also support doing some pipeline operations on the GPU and in a distributed fashion. Perhaps it's worth having a good look. They did the work for the Jasper ASR, so they are aware of the speech recognition context as well.

Thanks, I will look at it.

whisper-yu (approved):

LIZHICHAOUNICORN (approved):

PR #627, opened 2021-09-18
[examples] add shard tools and example on aishell

robin1001:
This is the shard tools and example on aishell to support the unify io, see https://github.com/wenet-e2e/wenet/pull/620


robin1001:
This is the aishell example of unified IO, see https://github.com/wenet-e2e/wenet/pull/627.

LIZHICHAOUNICORN (approved):

whisper-yu (approved):

PR #637, opened 2021-09-26
Support wenet x86 cmake build being a sub-project

daanzu:


robin1001:
Could you give a brief description about what's this PR for?

daanzu:
@robin1001 Oops, sure. This PR makes it easy to use the WeNet x86 runtime as a sub-project in a CMake build, with another project as the top-level project. Without this PR, the WeNet x86 runtime CMakeLists assumes it is in the root directory of the build, which causes build problems when it is actually in a sub-directory of the top-level build.

We also make the directories for WeNet's openfst build available to other (top-level) CMake projects, so they can link to the same openfst libraries.

By the way, thanks for the great project!

robin1001:
I see, great!

robin1001 (approved):

PR #643, opened 2021-09-27
[network] deprecated old io

robin1001:
Change-Id: Ib9276daa648d42ee86e5176d231ca66aae8e1e9d


qmpzzpmq:
it would be better to add a warning in 'tools/compute_cmvn_stats_deprecated.py ' when the script start like:
'this script will be removed later on' or something else

robin1001:
> it would be better to add a warning in 'tools/*_deprecated.py ' when the script start like:
> 'this script will be removed later on' or something else

Good suggestion, added!

qmpzzpmq (approved):
Great work

pengzhendong (approved):

PR #679, opened 2021-10-13
added implementation of compact-T.fst

JPalakapilly:
https://arxiv.org/pdf/2110.03098.pdf describes various WFST topologies for decoding with CTC.

The current `ctc_token_fst.py` creates the Eesen T.fst described in the paper. 
`ctc_token_fst_corrected.py` creates the "correct" T.fst.

This PR adds an implementation of the compact T.fst described in the paper.

The compact T.fst ends up with much smaller TLGs with virtually identical decoding performance. In our use case, a TLG that was initially 35MB when compiled with `ctc_token_fst.py` is only 24MB when compiled with `ctc_token_fst_compact.py`.

### DIFF
--- a/tools/fst/ctc_token_fst_compact.py
+++ b/tools/fst/ctc_token_fst_compact.py
@@ -1,0 +1,21 @@
+#!/usr/bin/env python
+
+import sys
+
+print('0 0 <blank> <eps>')
+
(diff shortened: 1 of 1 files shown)


JPalakapilly:
I have not run it on any public benchmarks. 
in our tests we saw tiny improvement like 13.22% -> 13.21% WER on an internal benchmark. almost identical results.

PR #686, opened 2021-10-15
fix context bug

peggyxpxu:
fix some bugs in add context for ctc prefix beam search.
1、context_state should be initialized to zero.
2、when context_score < 0, link_extra_cost should add link->context_score to avoid all links being deleted.


robin1001:
Good catch. typo error? from the code,  all the fix are related to WFST search, not prefix beam search.
 Please fix the lint error, and any test result before and after this fix?

robin1001:
Just follow the checks' link to find where you are wrong.

PR #709, opened 2021-10-22
Overrides arg

qmpzzpmq:
In calling of train.py/recognize.py, the extrac command args could be used for override the parameter in config.yaml
like
`
optim_conf:
    lr: 0.01
`
if the command
`python wenet/bin/train.py --override_config 'optim_conf.lr 0.001'`
is running, the decoding.ctc_weight will be override to 0.001.
like https://espnet.github.io/espnet/espnet2_training_option.html


zhuzizyf:
Fix override boolean args bug. 
image

Issue #745, opened 2021-10-29
tolower() used cannot properly lowercase accented letters


lushao1231:
**Describe the bug**
Wenet outputs capitalized accented letters:
e.g., yo recuerdos. no. estoy recordando. su unos cuatro, cinco aÑos atrÁs., me tocÓ! participar en una convocaciÓn. que habÍan mos hecho en mÉxico a fa

**To Reproduce**
Accented letters don't work in general. Sending spanish audio clips to wenet is one way to reproduce it.

**Expected behavior**
yo recuerdos. no. estoy recordando. su unos cuatro, cinco años atrás., me tocó! participar en una convocación. que habÍan mos hecho en méxico a fa

**Additional context**
a fix for this issue:

2 changes:

[ runtime/core/utils/string.cc ]
```cpp
#include <sstream>
#include <string>
#include <vector>

//**line 20-23**
#include "utils/log.h"
#include "utils/utils.h"
#include "utils/utf8_lower.h"

namespace wenet {
    //**line 150-157**
    // Ignore tailing space
    if (!result.empty() && result.back() == ' ') {
      result.pop_back();
    } 
    result = UTF8Lower(result);
  }
  return result;
}
```

[ runtime/server/x86/CMakeLists.txt ]
````cmake
//**line 81-85**
add_library(utils STATIC
  utils/string.cc
  utils/utils.cc
  utils/utf8_lower.cc
)

# The original openfst uses GNU Build System to run configure and build.
````
2 new files:

[ runtime/core/utils/utf8_lower.h ]
```c
#ifndef UTF8_LOWER_H_    
#define UTF8_LOWER_H_

#include <string>

unsigned char* StrToLwrExt(unsigned char* pString);

std::string UTF8Lower(std::string str);

#endif
```

[ runtime/core/utils/utf8_lower.cc ]
````cpp
#include "utils/utf8_lower.h"
#include <iostream>
#include <cctype>
#include <clocale>
#include <stdio.h>
#include <algorithm>
#include <string.h>

/*
    Function to lowercase a UTF-8 string, taken from:
      https://stackoverflow.com/questions/36897781/how-to-uppercase-lowercase-utf-8-characters-in-c
    Unclear if it's fully tested.
*/

unsigned char* StrToLwrExt(unsigned char* pString)
{
  unsigned char* p = pString;
  unsigned char* pExtChar = 0;

  if (pString && *pString) {
        while (*p) {
[...]

PR #749, opened 2021-10-29
[runtime] replace std::tolower() with boost::locale::to_lower()

xingchensong:
see issue https://github.com/wenet-e2e/wenet/issues/745


xingchensong:
we use wstring instead of boost::locale, see PR https://github.com/wenet-e2e/wenet/pull/754

PR #760, opened 2021-11-04
Fixed subprocess residue in UIO

zywang20:
Fixed subprocess residue in dataset process


wenet/dataset/processor.py
```diff
@@ -102,6 +104,7 @@ def tar_file_and_group(data):
             example['key'] = prev_prefix
             yield example
         stream.close()
+        sample['process'].communicate()
```
robin1001, line 107: if 'process' in sample: sample['process'].communicate()

robin1001 (approved):

PR #774, opened 2021-11-08
[runtime] fix duplicate output for kEndFeats

Mddct:
fix duplicate output for kEndFeats, because rescore will call  UpdateResult() again


robin1001:
> fix duplicate output for kEndFeats, because rescore will call UpdateResult() again

We use different parameters in the two cases for UpdateResult()

Mddct:
if (DecodedSomething()) {
    VLOG(1) << "Partial CTC result " << result_[0].sentence;
  }

robin1001:
Yes, it should be twice. For searching with LM, the results of  the two may be different.

PR #786, opened 2021-11-14
GPU infer

Slyne:
Perf result is not uploaded.
Will add later.


runtime/server/x86_gpu/scripts/convert.py
```diff
@@ -0,0 +1,100 @@
...
+        with open(os.path.join(model_dir, template), "r", encoding="utf-8") as f:
+            for line in f:
+                if line.startswith("#"):
+                    continue
+                for key, value in model_params.items():
+                    if key == "#encoder_out_feat_size" and \
+                       configs["encoder_conf"].get("causal", False) and \
+                       model == "encoder":
+                        value = "-1"
+                    line = line.replace(key, str(value))
```
xingchensong, line 89: plz add some comments for this special if-branch, why encoder_out_feat_size should be -1 if encoder uses causal conv? If so , we will get shape [-1, -1] in `model_repo/encoder/config_template.pbtxt::line38`, is it possible for onnx or triton to handle such a strange shape?

runtime/server/x86_gpu/model_repo/scoring/1/model.py
```diff
@@ -0,0 +1,287 @@
...
+                                                   min(total, self.num_processes),
+                                                   blank_id=self.blank_id,
+                                                   space_id=-2,
+                                                   cutoff_prob=self.cutoff_prob,
+                                                   ext_scorer=self.lm)
+        all_hyps = []
+        all_ctc_score = []
+        max_seq_len = 0
+        for seq_cand in score_hyps:
+            # if candidates lenss than beam size
```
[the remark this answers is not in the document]
Slyne: yes, typo.

runtime/server/x86_gpu/scripts/convert.py
Slyne: It seems very strange that the exported model including those transformers and offline conformers all have a -1 (dynamic) output size. We need to take a close look at this issue. But it doesn't affect deployment.

xingchensong: i c, i think we'd better add some notes for this if-branch to let users know the reason for doing this and leave the code as it is. could u add some comments like [this](https://github.com/wenet-e2e/wenet/pull/786/files#diff-c1fac5c7b9cb061736a33b160029d301823b2d52e2d9f1c74f733d2de54d7d8fR135-R136)? Thx.

Slyne: Sure.

xingchensong:
@Slyne hi, do you prefer to add perf result in another PR? if so, I will merge this one, many thx!

Slyne:
@xingchensong Just updated the perf. Thanks!

xingchensong (approved):

PR #809, opened 2021-11-22
layer reuse

qmpzzpmq:


jprobichaud:
This looks really interesting, it opens a bunch of possibilities going forward.  Any chances we can get this merged anytime soon?

qmpzzpmq:
it actually mainly design for pack model into .apk.

jprobichaud:
I've applied this PR and trained with my data. This works well in the python world (recognize.py) but trying it on the c++ interface with decoder_main doesn't: i get barely any words out.  

Is there something special to do in the c++ code to leverage this? I see in the zip archive thatbthe for-loop for the repeated layer is indeed there but could it be that we need something else in the C++ code?

qmpzzpmq:
@jprobichaud Would you mind post the WER result after you use this PR?
And, Can I see your C++ code? Maybe we can work together for this.

jprobichaud:
> Would you mind post the WER result after you use this PR?
99.6% WER : mostly all deletions...

We have custom code to call the wenet structure, but it's mostly what decoder main produces...  Decoding with decoder_main exibit the same issue.

qmpzzpmq:
If you directly depoly it on runtime, I think so. But would you mind post the original WER and after using the WER using recognize.py?

jprobichaud:
> If you directly depoly it on runtime, I think so. But would you mind post the original WER and after using the WER using recognize.py?

I am not testing on public data sets, so reporting the WER won't be super useful : all test suites I try yield the same issue : all deletions.  My dataset isn't composed of short utterances, so I never use recognize.py to assest accuracy if an entire test suite, but only to debug or inspect what happens on one shorter utterance.

Maybe that could satisfy you:

using recognize.py with the layer reuse (not trained fully), a short utt produced the following:
```
2021-12-23 15:54:28,223 INFO PIPER_DEVER_00000 ▁yes▁he▁rains▁yes▁the▁king▁and▁the▁kingdom▁have▁come▁but▁no▁he▁iss▁not▁good▁now
```
The "corresponding" model (same number of layers, same tokenizer, but without any layer reuse) produces:
```
2021-12-23 16:13:25,931 INFO PIPER_DEVER_00000 ▁yes▁he▁reigns▁yes▁the▁king▁and▁the▁kingdom▁have▁come▁but▁no▁that▁is▁not▁good▁news
```
And this corresponds exactly to the content of the audio.

If I use decode_main, I get nothing out, if I use the traditional model, I get the right output.

jprobichaud:
Note that the layer-reuse model wasn't trained for the same numbers of epochs, which could explain some accuracy issues.   But before investing in training it for a long time, I need to make sure I can use it on the runtime. 

My gut feeling is that the repetition of layer isn't happening correctly somehow inside the jit version.

qmpzzpmq:
the reason I asked for recognize.py is I think somehow a problem happen in https://github.com/wenet-e2e/wenet/blob/main/wenet/transformer/encoder.py#L170.
And If you can test recognize.py, it would be better. If not, I will get back to check in 2022.

qmpzzpmq:
@jprobichaud it should be fixed. Please try and give me feedback

jprobichaud:
I get the following when I try to export the model:
```
RuntimeError: 
Module 'ConformerEncoder' has no attribute 'num_blocks' :
  File "/shared/workspaces/jp/NeMo/rev-wenet/examples/rev/s0/wenet/transformer/encoder.py", line 244
                    attn_cache = None
                else:
                    attn_cache = elayers_output_cache[self.num_blocks * j + i]
                                                      ~~~~~~~~~~~~~~~ <--- HERE
                    #attn_cache = elayers_output_cache[nblocks * j + i]
                if conformer_cnn_cache is None:
```
If I try to set `self.num_blocks = num_blocks` in any of the constructors, the model exports fine, but decoder_main halts with a segfault while loading the model

qmpzzpmq:
let me check

robin1001:
If it works, I think we can add another LayerReuseTransformer, and override the forward function. In this way, we will not affect the old code of standard transformer/conformer, and the code is also clean.

jprobichaud:
I see that this PR is closed with being merged, is the intention to redo it implementing a new LayerReuseTransformer ?

qmpzzpmq:
> I see that this PR is closed with being merged, is the intention to redo it implementing a new LayerReuseTransformer ?

@jprobichaud hi I have a idea to improve this lay-reuse tech. As you know, now with number of parameters significantly dropping but WER increase as well. I introduced a new tech could decrease WER (haven't get specific number) with slight maybe around ~5% parameters increase from layer-reuse transformer. But I haven't done the test.

Issue #796, opened 2021-11-17
processor.tokenize support non-lang-syms


yzmyyff:
We can assign --non-lang-syms in the old io framework, such as text2token.py and format_data. It leaves a special symbol([laugh] or [noise]) as it is.

But in uio we don't support this feature. 

Do we have any plan for this? Or how do I achieve the same purpose?

robin1001:
I didn't go through the details of  --non-lang-syms. do you have the  bandwidth to support that?

yzmyyff:
I'd like to but I'm not free recently. Maybe next week or after.

PR #819, opened 2021-11-25
Support non-linguistic symbol in uio

yzmyyff:
resolves https://github.com/wenet-e2e/wenet/issues/796

### DIFF
--- a/wenet/dataset/dataset.py
+++ b/wenet/dataset/dataset.py
@@ -120,7 +120,7 @@ def Dataset(data_type,
             data_list_file,
             symbol_table,
             conf,
-            bpe_model=None,
+            bpe_model=None, non_lang_syms=None,
             partition=True):
     """ Construct dataset from arguments
 
@@ -144,7 +144,7 @@ def Dataset(data_type,
         dataset = Processor(dataset, processor.parse_raw)
 
     dataset = Processor(dataset, processor.tokenize, symbol_table, bpe_model,
-                        conf.get('split_with_space', False))
+                        non_lang_syms, conf.get('split_with_space', False))
     filter_conf = conf.get('filter_conf', {})
     dataset = Processor(dataset, processor.filter, **filter_conf)
 
--- a/wenet/dataset/dataset.py
+++ b/wenet/dataset/dataset.py
@@ -120,7 +120,8 @@ def Dataset(data_type,
             data_list_file,
             symbol_table,
             conf,
-            bpe_model=None, non_lang_syms=None,
+            bpe_model=None,
+            non_lang_syms=None,
             partition=True):
     """ Construct dataset from arguments
 
--- a/wenet/dataset/processor.py
+++ b/wenet/dataset/processor.py
@@ -25,6 +25,8 @@ import torchaudio
 import torchaudio.compliance.kaldi as kaldi
 from torch.nn.utils.rnn import pad_sequence
 
+from tools.text2token import exist_or_not
+
 AUDIO_FORMAT_SETS = set(['flac', 'mp3', 'm4a', 'ogg', 'opus', 'wav', 'wma'])
 
 
@@ -282,7 +284,52 @@ def compute_fbank(data,
         yield dict(key=sample['key'], label=sample['label'], feat=mat)
 
 
-def tokenize(data, symbol_table, bpe_model=None, split_with_space=False):
+def __tokenize_non_lang_syms(txt, non_lang_sym_patterns):
+    """split txt by non_lang_sym_patterns
+
+    Args:
+        txt: original transcript.
+        non_lang_sym_patterns: non lang sym regex pattern.
+
+    Returns:
+        [token, token, ..., token]
+
+    """
+    match_pos = []
+
+    for r in non_lang_sym_patterns:
+        i = 0
+
+        while i >= 0:
+            m = r.search(txt, i)
+
+            if m:
+                match_pos.append([m.start(), m.end()])
+                i = m.end()
+            else:
+                break
+
+    if len(match_pos) > 0:
+        chars = []
+        i = 0
+
+        while i < len(txt):
+            start_pos, end_pos = exist_or_not(i, match_pos)
+
+            if start_pos is not None:
+                chars.append(txt[start_pos: end_pos])
+                i = end_pos
+            else:
+                chars.append(txt[i])
+                i += 1
+
+        return chars
+    else:
+        return txt
+
+
+def tokenize(data, symbol_table, bpe_model=None, non_lang_syms=None,
+             split_with_space=False):
     """ Decode text to chars or BPE
         Inplace operation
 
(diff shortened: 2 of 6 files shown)


wenet/dataset/dataset.py
```diff
@@ -120,7 +120,7 @@ def Dataset(data_type,
             data_list_file,
             symbol_table,
             conf,
-            bpe_model=None,
+            bpe_model=None, non_lang_syms=None,
```
robin1001, line 123: newline and align.

wenet/dataset/processor.py
```diff
@@ -292,13 +339,20 @@ def tokenize(data, symbol_table, bpe_model=None, split_with_space=False):
...
+
     if bpe_model is not None:
         import sentencepiece as spm
         sp = spm.SentencePieceProcessor()
         sp.load(bpe_model)
+
     for sample in data:
         assert 'txt' in sample
         txt = sample['txt'].strip()
+        txt = __tokenize_non_lang_syms(txt, rs)
```
robin1001, line 355:
``` python
if len(rs) > 0:
    __tokenize_non_lang_syms(txt, rs)
```

wenet/dataset/processor.py
```diff
@@ -282,7 +284,52 @@ def compute_fbank(data,
...
+
+    Returns:
+        [token, token, ..., token]
+
+    """
+    match_pos = []
+
+    for r in non_lang_sym_patterns:
+        i = 0
+
```
robin1001, line 302: remove empty line

wenet/dataset/processor.py
```diff
@@ -282,7 +284,52 @@ def compute_fbank(data,
...
+        while i >= 0:
+            m = r.search(txt, i)
+
```
robin1001, line 305: remove empty line.

wenet/dataset/processor.py
yzmyyff: This function just returns input txt if rs is an empty list

robin1001: Please also take bpe mode into consideration into account. The code fails for BPE.

robin1001: > 

we should avoid to call the function to save time especially for very big data.

xingchensong:
Hi, thanks for your great contribution, however, i do think we can make the impl much simpler. For example, If we force that non-linguistic symbols must appear in a fixed format (i.e. `{XXXXXX}`), then function `__tokenize_non_lang_syms` can be simply replaced with three lines of code:
```py
non_lang_syms_pattern = re.compile(r'(\{[A-Z]+\})')
pieces = non_lang_syms_pattern.split(txt.upper())
pieces = [w for w in pieces if len(w.strip()) > 0]
```

below is the verification:
```py
import re

txt = "我是{NOISE} do you known im{AAA} haha {nnb}s {c}"

# Solution-a
non_lang_syms = ["{NOISE}", "{AAA}", "{nnb}", "{c}"]
rs = [re.compile(re.escape(x)) for x in non_lang_syms]


def exist_or_not(i, match_pos):
    start_pos = None
    end_pos = None
    for pos in match_pos:
        if pos[0] <= i < pos[1]:
            start_pos = pos[0]
            end_pos = pos[1]
            break

    return start_pos, end_pos


def __tokenize_non_lang_syms(txt, non_lang_sym_patterns):
    """split txt by non_lang_sym_patterns
    Args:
        txt: original transcript.
        non_lang_sym_patterns: non lang sym regex pattern.
    Returns:
        [[piece, is_non_lang_sym], [piece, True], ..., [piece, False]]
    """
    match_pos = []
    for r in non_lang_sym_patterns:
        i = 0
        while i >= 0:
            m = r.search(txt, i)
            if m:
                match_pos.append([m.start(), m.end()])
                i = m.end()
            else:
                break

    if len(match_pos) > 0:
        pieces = []
        i = 0
        while i < len(txt):
            start_pos, end_pos = exist_or_not(i, match_pos)
            if start_pos is not None:
                pieces.append([txt[start_pos: end_pos], True])
                i = end_pos
            else:
                if len(pieces) == 0 or (len(pieces) > 0 and pieces[-1][1]):
                    pieces.append([txt[i], False])
                else:
                    pieces[-1][0] += txt[i]
                i += 1

        return pieces
    else:
        return [[txt, False]]


parts_a = __tokenize_non_lang_syms(txt, rs)

# Solution-b
non_lang_syms_pattern = re.compile(r'(\{[A-Z]+\})')
parts_b = non_lang_syms_pattern.split(txt.upper())
parts_b = [w for w in parts_b if len(w.strip()) > 0]


# Check
assert len(parts_a) == len(parts_b)
for i, j in zip(parts_a, parts_b):
    assert i[0].strip().upper() == j.strip().upper()
```

Besides, we can also simplify the control-flow of function `tokenize()` by reusing the BPE/non-BPE branch without any modification:
```py
non_lang_syms_pattern = re.compile(r'(\{[A-Z]+\})')  # fixed format
for sample in data:
    assert 'txt' in sample
    txt = sample['txt'].strip()
    parts = non_lang_syms_pattern.split(txt.upper())
    parts = [w for w in parts_b if len(w.strip()) > 0]
    for part in parts:
        if part in non_lang_syms:
            tokens.append(part)
        else:
            assert part[0] != '{'  # do some checks
            if bpe_model is not None:
                 reuse BPE logic
            else:
                reuse non-BPE logic
```

yzmyyff:
In my own data non-lang symbols are all surrounded by "{".
So it's OK for me to use the regex split.

In some data, the format of the non-lang symbols is mixed. So we need to construct a complicated pattern. (eg. <noise>, {BRK}, #, ## and so on)

And most parts of this commit are copied from text2token.py. I am not sure if there is a potential purpose...

xingchensong:
We can just add some format check after reading the non-ling symbol table 
https://github.com/wenet-e2e/wenet/pull/819/files#diff-a5e6ec48421b12c254127394ff8947e243de71257d2d7b4868dc9fdcb06adb4eR42

if it contains `#` or something else， raise errors and give some suggestions within error messages: "non-ling symbol should be formatted in {XXXX}, consider change # to {SHARP}".

the reasons for choosing fixed format lies in:
- code is easier to read and understand
- the number of non-ling symbols is small, it is also easy for users to define and diy their own non-ling symbol table

robin1001:
@yzmyyff Seems xingchen's solution is more simple, could you refine it according to his suggestion?

yzmyyff:
In test code, If we change non_lang_syms to ```["{NOISE}", "{AAA}", "{nnb}"]```. These two method has different result.

We must construct a more complicated regex pattern to cover this. I think we just move the complexity to the construction stage.

xingchensong:
> In test code, If we change non_lang_syms to `["{NOISE}", "{AAA}", "{nnb}"]`. These two method has different result.
> 
> We must construct a more complicated regex pattern to cover this. I think we just move the complexity to the construction stage.

1. leaving `{c}` as a normal linguistic symbol is kind of weird
2. we can double check non_lang_syms in tokenize(), see:

image

yzmyyff:
Got it. Thanks for your suggestion!

robin1001:
We have to take non-liguistic symbol pattern into account. for example, the pattern is:
[xxx] for swithboard and fisher
<xxx> for wsj and chime4

robin1001:
I mean we should take `{xxx}` `<xxx>`  `[xxx]`  into account.

yzmyyff:
As I mentioned before, If we do not consider forward compatibility and add some documentation about the special symbols. The new method would be a good starting.

Otherwise, the previous version may be simpler.

robin1001:
Yes, we see. xingchen just want to keep it simple and stupid, and the pattern is just for some english dataset which has special symbols, and the above three patterns are enough to cover.

yzmyyff:
In fact, I have seen some corpus use "#" to indicate the length of speech pause. # means stop level 1, ## means stop level 2...

I think leaving it to the user is a good choice.

robin1001:
#1/#2/#3 is for TTS corpus. here we only consider the public speech recognition dataset so it's convenient for researchers.
This symbols are meaningless in production speech dataset.

yzmyyff:
Alright

robin1001 (approved):

PR #829, opened 2021-12-02
[examples] add wsj recipe

Shengqiang-Li:


examples/wsj/s0/postprocess.py
```diff
@@ -0,0 +1,49 @@
+import os
```
robin1001, line 1: what is the file for? please put it to local.

Issue #864, opened 2021-12-16
the context biasing that has not yet the end boundary does not subtract the previous offset score.


luchuanze:
**Describe the issue**

At the end of the ctc prefix search, the context that has not yet the end boundary does not subtract the previous offset score.

**To Reproduce**

My test context is "千玺机器人" ，context_score= 3.0, and my test pronouncing is "qianxi" .  The recognition results are as follows：

![image](https://user-images.githubusercontent.com/32317033/146340351-3e2e73d4-de88-4057-9b4a-301d5c8576a3.png)

When the ctc search was over,  a relatively low score for "迁徙" because the context state of "千玺" was not set to the start state with negative score.

**Expected behavior**

By overwritting 'CtcPrefixBeamSearch.FinalizeSearch()' to fix it ?

robin1001:
@pengzhendong , what do you think?

pengzhendong:
@luchuanze We've talked about this before. You could commit a pr, thanks!

luchuanze:
> @luchuanze我们以前讨论过这个。你可以提交一个公关，谢谢！

luchuanze:
> @luchuanze We've talked about this before. You could commit a pr, thanks!

OK,

PR #872, opened 2021-12-18
Handle the context that has not yet the end boundary does not set to start state with negative score 

luchuanze:
Handle the context that has not yet the end boundary does not set to start state with negative score at the end of ctc prefix search at the end of ctc prefix search.
See the issue #864.

### DIFF
--- a/runtime/core/decoder/ctc_prefix_beam_search.cc
+++ b/runtime/core/decoder/ctc_prefix_beam_search.cc
@@ -186,4 +186,46 @@ void CtcPrefixBeamSearch::Search(const torch::Tensor& logp) {
   }
 }
 
+void CtcPrefixBeamSearch::FinalizeSearch()
+{ 
+  //TODO(luchuanze)
+  UpdateFinalContext();
+}
+
+void CtcPrefixBeamSearch::UpdateFinalContext() {
+  if (context_graph_ == nullptr) return;
+
+  CHECK_EQ(hypotheses_.size(), cur_hyps_.size());
+  CHECK_EQ(hypotheses_.size(), likelihood_.size());
+  // TODO(luchuanze)
+  // The context that has not yet the end boudary does not set to start state
+  // with negative score, so using a word id -1 to udpate the hyps itself at the
+  // end of the search.
+  for (const auto& prefix : hypotheses_) {
+    PrefixScore& prefix_score = cur_hyps_[prefix];
+    prefix_score.UpdateContext(context_graph_, prefix_score, -1, prefix.size());
+  }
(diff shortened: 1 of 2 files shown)


runtime/core/decoder/ctc_prefix_beam_search.cc
```diff
@@ -186,4 +186,42 @@ void CtcPrefixBeamSearch::Search(const torch::Tensor& logp) {
...
 
+void CtcPrefixBeamSearch::FinalizeSearch() {
+  UpdateFinalContext();
+}
+
+void CtcPrefixBeamSearch::UpdateFinalContext() {
+  if (context_graph_ == nullptr) return;
+  CHECK_EQ(hypotheses_.size(), cur_hyps_.size());
+  CHECK_EQ(hypotheses_.size(), likelihood_.size());
+  // The context that has not yet the end boudary does not set to start state
```
robin1001, line 197: Please refinte the comment.
We should backoff the context score/state when the context is not fully matched at the last frame.

runtime/core/decoder/ctc_prefix_beam_search.cc
```diff
@@ -186,4 +186,42 @@ void CtcPrefixBeamSearch::Search(const torch::Tensor& logp) {
...
+  // with negative score, so using a word id -1 to udpate the hyps itself at the
```
robin1001, line 198: use 0 instead of -1.

runtime/core/decoder/ctc_prefix_beam_search.cc
```diff
@@ -186,4 +186,42 @@ void CtcPrefixBeamSearch::Search(const torch::Tensor& logp) {
...
+    PrefixScore& prefix_score = cur_hyps_[prefix];
+    prefix_score.UpdateContext(context_graph_, prefix_score, -1, prefix.size());
+  }
+
+  // Resort hyps
+  std::vector<std::pair<std::vector<int>, PrefixScore>> arr(cur_hyps_.begin(),
+                                                              cur_hyps_.end());
+  std::sort(arr.begin(), arr.end(), PrefixScoreCompare);
+
+  // Update cur_hyps_ and get new result
```
robin1001, line 210: The following code is repetitive with what in `Search`, please try to refine.

luchuanze: OK, I will refine it.

luchuanze:
@robin1001 Hi Binbin,. I have refined the code, please review it.

pengzhendong (approved):

Issue #842, opened 2021-12-07
Use DDP training to get stuck


601222543:
## Describe the bug
I got stuck when using DDP training with my own wenet and my own data. And stuck(GPU utilization 100%) at the beginning of the second epoch every time.
After debugging, it was found to be stuck in this position:
```
# wenet/utils/executor.py
with torch.cuda.amp.autocast(scaler is not None):
    loss, loss_att, loss_ctc = model(
        feats, feats_lengths, target, target_lengths)
```

## Environment
CentOS Linux release 7.8.2003 (Core)
GPU Driver Version: 450.80.02
CUDA Version: 10.2
torch==1.8.0
torchaudio==1.8.1
torchvision==0.9.0

## Some Attempts
I did some attempts later and found:
1 gpu no problem
multi gpu stuck
`static batch` no problem
`dynamic batch` stuck
`conformer` no problem
`unified_conformer` stuck

Other attempts：
Upgrade pytorch version to 1.9.0, 1.10.0 is useless
Set num_workers=0/1 is useless
V100 -> P40 useless
Sleep 1 minute after completing an epoch is useless
NCCL is completely stuck without error log
GLOO error log:
```
2021-12-07 11:36:17,011 INFO Epoch 0 CV info cv_loss 115.3632936241356
2021-12-07 11:36:17,011 INFO Epoch 1 TRAIN info lr 6.08e-06
2021-12-07 11:36:17,014 INFO using accumulate grad, new batch size is 8 times larger than before
2021-12-07 11:36:17,335 INFO Epoch 0 CV info cv_loss 115.36239801458647
2021-12-07 11:36:17,335 INFO Epoch 1 TRAIN info lr 6.200000000000001e-06
2021-12-07 11:36:17,338 INFO using accumulate grad, new batch size is 8 times larger than before
2021-12-07 11:36:17,579 INFO Epoch 0 CV info cv_loss 115.36309641650827
2021-12-07 11:36:17,579 INFO Epoch 1 TRAIN info lr 5.96e-06
2021-12-07 11:36:17,582 INFO using accumulate grad, new batch size is 8 times larger than before
2021-12-07 11:36:17,926 INFO Epoch 0 CV info cv_loss 115.36275817930736
2021-12-07 11:36:17,926 INFO Checkpoint: save to checkpoint exp/conformer/0.pt
2021-12-07 11:36:18,889 INFO Epoch 1 TRAIN info lr 6.32e-06
[...]

PR #880, opened 2021-12-23
try to solve the programm stuck, when train and cv both are dynamic batch size

zelda3721:
try to solve the programm stuck, when train and cv both are dynamic batch size,according to the issue 
https://github.com/wenet-e2e/wenet/issues/842#issue-1072991518

### DIFF
--- a/wenet/utils/executor.py
+++ b/wenet/utils/executor.py
@@ -40,7 +40,8 @@ class Executor:
         with model_context():
             iterator_stop = torch.tensor(0).to("cuda")
             for batch_idx, batch in enumerate(data_loader):
-                # try to solve the programm stuck, when train and cv both are dynamic batch size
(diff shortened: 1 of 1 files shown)


robin1001:
Why it is stuck for dynamic batch size?

zelda3721:
> 

Mybe it is because the iteration counts are not always equals between processes, and why this happened may because the data is not balance. 
In my situation, I used my own data  with many short words, and trained with mult_cn dataset. When the stuck happened, there are also many defunct processes.

robin1001:
Acctually, we use model.join() to process uneven data, please see https://pytorch.org/tutorials/advanced/generic_join.html

robin1001:
Fixed in a simple way, please see https://github.com/wenet-e2e/wenet/pull/903/

PR #881, opened 2021-12-23
Support loading of the  pretrained conformer encoder 

Emiyassstar:
Support loading  encoder mods
Support swap the order of the convolution module and the multi-head self-attention module
add CNN feature norm


robin1001:
Sorry, I missed the PR. Seems the PR contains the following part:
1. openasr2021 recipe
2. loadding the pretrained model
3. conformer change

It's better to separate it in three PRs to keep each pr simple and independent.

PR #896, opened 2021-12-31
Add a link to open source onnxruntime version

znsoftm:


robin1001:
onnx is not a platform or device, so the pr is rejected.

PR #907, opened 2022-01-09
change layer norm to use default eps

Slyne:
Signed-off-by: slyned <slyned@nvidia.com>

NAN will occur when using model.half() to do inference.
The issue is reported here: https://github.com/pytorch/pytorch/issues/41527

1. Changed all eps=1e-12 to 1e-5 a default value for [torch.nn.functional.layer_norm](https://pytorch.org/docs/stable/generated/torch.nn.LayerNorm.html).

### DIFF
--- a/wenet/transformer/decoder.py
+++ b/wenet/transformer/decoder.py
@@ -63,7 +63,7 @@ class TransformerDecoder(torch.nn.Module):
             raise ValueError(f"only 'embed' is supported: {input_layer}")
 
         self.normalize_before = normalize_before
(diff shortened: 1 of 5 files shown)


robin1001:
Any WER number of the change?  You can test it on a pretrained model.

Slyne:
> Any WER number of the change? You can test it on a pretrained model.

Tested on Aishell2 pretrained model, ~0.02% CER increase on ctc prefix & ctc greedy, ~0.01%CER increase on attention rescoring. Let me test more pretrained models.

robin1001 (approved):

PR #921, opened 2022-01-18
[runtime] add onnxruntime export and inference

lzhin:
We have implemented the derivation and reasoning of wenet onnxruntime.
On the aishell test dataset, the CER of libtorch and onnx is 2.2, the RTF of libtorch is 0.1379, and the RTF of onnx is 0.1166, which is relatively reduced by 15.45%.

### DIFF
--- a/runtime/core/bin/decoder_main_onnx.cc
+++ b/runtime/core/bin/decoder_main_onnx.cc
@@ -1,0 +1,139 @@
+// Copyright 2020 Mobvoi Inc. All Rights Reserved.
+// Author: binbinzhang@mobvoi.com (Binbin Zhang)
+//         di.wu@mobvoi.com (Di Wu)
+// Copyright 2021 Huya Inc. All Rights Reserved.
+// Author: lizexuan@huya.com (Zexuan Li)
+
+#include <iomanip>
+#include <utility>
+
+#include "torch/script.h"
+
+#include "decoder/params.h"
+#include "frontend/wav.h"
+#include "utils/flags.h"
+#include "utils/log.h"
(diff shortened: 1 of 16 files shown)


robin1001:
Great, quite a lot of code, it takes time to review it. However, I will do it as soon as possible.

robin1001:
可以加我微信 binbzha，我拉你进和几个做这块的同学讨论。

jprobichaud:
This is very interesting.  Any number on the RTF in either streaming or offline mode compared to the current torch.jit approach on CPU?

lzhin:
> This is very interesting. Any number on the RTF in either streaming or offline mode compared to the current torch.jit approach on CPU?

In streaming mode, the RTF of libtorch is 0.1379, and the RTF of onnx is 0.1166.

jprobichaud:
> > This is very interesting. Any number on the RTF in either streaming or offline mode compared to the current torch.jit approach on CPU?
> 
> In streaming mode, the RTF of libtorch is 0.1379, and the RTF of onnx is 0.1166.

Thanks a lot!

Mddct:
There are three sessions, each session has own thread pool acorrding https://onnxruntime.ai/docs/get-started/with-c.html.

lzhin:
> There are three sessions, each session has own thread pool acorrding https://onnxruntime.ai/docs/get-started/with-c.html. Should we use "Global/shared threadpools" between sessions or export onnx just a model

Thanks for your advice. I'll try to upgrade the code when I'm free.

xingchensong:
Closed, we refine our code in PR #1023  #1103

PR #924, opened 2022-01-20
Fixed an issue where some examples might abort

chereyk:
The script in examples will stop reporting errors because no folder is created, which may cause some trouble for students who are not familiar with scripting languages

### DIFF
--- a/examples/aishell/s0/local/download_and_untar.sh
+++ b/examples/aishell/s0/local/download_and_untar.sh
@@ -23,8 +23,8 @@ url=$2
 part=$3
 
 if [ ! -d "$data" ]; then
-  echo "$0: no such directory $data"
-  exit 1;
+  echo "$0: no such directory $data, make it"
+  mkdir -p $data
 fi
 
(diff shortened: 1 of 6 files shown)


robin1001:
This is not the right way to solve what you called `trouble`. You can control it from the caller script by setting 
``` bash 
set -e  
```
please see https://www.gnu.org/software/bash/manual/html_node/The-Set-Builtin.html.

chereyk:
sorry,i am not express clear, i mean
When I run bash run.sh --stage -1 --stop-stage -1 according to tutorial_aishell.md, there will be this error
stage -1: Data Download
local/download_and_untar.sh: no such directory /export/data/asr-data/OpenSLR/33 /
local/download_and_untar.sh: no such directory /export/data/asr-data/OpenSLR/33/
At this point, only need to add a command to create a folder to download normally

robin1001:
The right way is to change /export/data/asr-data/OpenSLR/33 to a directory which you have full access.

PR #927, opened 2022-01-24
conformer encoder modification

Emiyassstar:
We have some model modifications to support the pre-trained conformer encoder
1.Add CNN feature layer norm 
1.Support swap the order of the convolution module and the multi-head self-attention module
  Reference from  https://arxiv.org/abs/2011.10798


robin1001:
This is not going to be accepted.

PR #942, opened 2022-02-11
[Draft] Cache control v2

xingchensong:
This is not a merge-ready PR, I just push my testing code for discussion and further evaluation (such as GPU perf, ONNX export, ...). 

Performance on CPU (intel i7-10510U @ 1.80GHz), RTF from 0.1 -> 0.07, about 30% improvement:
image

Detailed descriptions (in Chinese): https://horizonrobotics.feishu.cn/sheets/shtcniLh77AgP6NJAXhd5UHXDwh

Test code:
```sh
bash rtf.sh --api 1 > log.txt.1
bash rtf.sh --api 2 > log.txt.2
grep "RTF:" log.txt.1
grep "RTF:" log.txt.2
```
u2++_conformer.zip: https://horizonrobotics.feishu.cn/file/boxcnO50Ea8m0rR2p9FwJ8ZHEIc
words.txt: https://horizonrobotics.feishu.cn/file/boxcnBpSEOWoBSIgLdlHetsjOFd

### DIFF
--- a/runtime/core/decoder/torch_asr_decoder.cc
+++ b/runtime/core/decoder/torch_asr_decoder.cc
@@ -38,15 +38,21 @@ TorchAsrDecoder::TorchAsrDecoder(
                                           resource->context_graph));
   }
   ctc_endpointer_->frame_shift_in_ms(frame_shift_in_ms());
+  infer_control_[3] = opts_.chunk_size;
 }
 
 void TorchAsrDecoder::Reset() {
   start_ = false;
   result_.clear();
   offset_ = 0;
+  std::fill(infer_control_.begin(), infer_control_.end(), 0);
+  infer_control_[3] = opts_.chunk_size;
   num_frames_ = 0;
   global_frame_offset_ = 0;
   num_frames_in_current_chunk_ = 0;
+  cnn_cache_ = std::move(torch::zeros({12, 0, 0, 0}));
+  key_cache_ = std::move(torch::zeros({12, 0, 0, 0, 0}));
+  value_cache_ = std::move(torch::zeros({12, 0, 0, 0, 0}));
   subsampling_cache_ = std::move(torch::jit::IValue());
   elayers_output_cache_ = std::move(torch::jit::IValue());
   conformer_cnn_cache_ = std::move(torch::jit::IValue());
@@ -62,7 +68,12 @@ void TorchAsrDecoder::ResetContinuousDecoding() {
   start_ = false;
   result_.clear();
   offset_ = 0;
+  std::fill(infer_control_.begin(), infer_control_.end(), 0);
+  infer_control_[3] = opts_.chunk_size;
   num_frames_in_current_chunk_ = 0;
+  cnn_cache_ = std::move(torch::zeros({12, 0, 0, 0}));
+  key_cache_ = std::move(torch::zeros({12, 0, 0, 0, 0}));
+  value_cache_ = std::move(torch::zeros({12, 0, 0, 0, 0}));
   subsampling_cache_ = std::move(torch::jit::IValue());
   elayers_output_cache_ = std::move(torch::jit::IValue());
   conformer_cnn_cache_ = std::move(torch::jit::IValue());
@@ -104,6 +115,14 @@ DecodeState TorchAsrDecoder::AdvanceDecoding() {
   // If not okay, that means we reach the end of the input
   if (!feature_pipeline_->Read(num_requried_frames, &chunk_feats)) {
     state = DecodeState::kEndFeats;
+    if (num_requried_frames < std::numeric_limits<int>::max() &&
+        num_requried_frames > chunk_feats.size()) {
+      // NOTE(xcsong): Padding last chunk to meet requirements of chunk_size
+      int pad_len = num_requried_frames - chunk_feats.size();
+      std::vector<std::vector<float>> pad_data(
+        pad_len, std::vector<float>(feature_dim, 0.0));
+      chunk_feats.insert(chunk_feats.end(), pad_data.begin(), pad_data.end());
+    }
   }
 
   num_frames_in_current_chunk_ = chunk_feats.size();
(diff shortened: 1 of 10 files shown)


xingchensong:
Please note that `forward_v2()` functions will be merged with `forward()` when all tests are done. Currently, we just keep them separate for easier comparison.

wenet/transformer/encoder.py
```diff
@@ -250,6 +250,58 @@ def forward_chunk(
...
+        assert cache_size == _cache_size
+        assert chunk_size == -1 or chunk_size == _chunk_size
+        xs = self.embed.out(
+            xs.transpose(1, 2).contiguous().view(b, _chunk_size, channel * f))
+        # Stage-3: PositionalEncoding
+        xs, _ = self.embed.pos_enc(xs, offset)
+        conformer_only_pos_emb = self.embed.position_encoding(
+            offset - cache_size, cache_size + _chunk_size)
+        # Stage-4: Encoder layers
+        mask_pad = torch.ones((0, 0, 0), device=xs.device, dtype=torch.bool)
```
robin1001, line 282: Using empty tensor instead of None, which is preferred for onnx export?

runtime/core/decoder/torch_asr_decoder.h
```diff
@@ -140,6 +141,19 @@ class TorchAsrDecoder {
   std::vector<std::vector<float>> cached_feature_;
   bool start_ = false;
 
+  // NOTE(xcsong): In order to simplify the comparison of different APIs,
+  //  all the data required by the two APIs are retained.
+
+  // API_v2
+  // cnn_cache_: (elayers, batch, cnn_kernel - 1, hidden_dim)
+  torch::Tensor cnn_cache_ = torch::zeros({12, 0, 0, 0});
```
robin1001, line 149: 12 is hard code here?

xingchensong:
hope so, need to be verified.

xingchensong:
Yep, the layer for-loop in forward_chunk_v2 use layer index to extract cache, thus the first dimension of cache tensor cannot be zero. We can just hardcord a large elayers (init cache tensor with shape [99,0,0,0,0]) to keep generalizability to other encoder later configs.

wenet/transformer/encoder.py
```diff
@@ -250,6 +250,58 @@ def forward_chunk(
         return (xs[:, cache_size:, :], r_subsampling_cache,
                 r_elayers_output_cache, r_conformer_cnn_cache)
 
+    def forward_chunk_v2(
+        self,
+        xs: torch.Tensor,
+        infer_control: torch.Tensor,
+        cnn_cache: torch.Tensor,
+        key_cache: torch.Tensor,
```
robin1001, line 258: seems key_cache and value cache shares the same operation and interface, could they be combined?

wenet/transformer/encoder.py
```diff
@@ -250,6 +250,58 @@ def forward_chunk(
...
+        if self.global_cmvn is not None:
+            xs = self.global_cmvn(xs)
+        # Stage-2: Subsampling
+        xs = xs.unsqueeze(1)  # (b, c1=1, frames, mel_bins)
+        xs = self.embed.conv(xs)
+        b, channel, _chunk_size, f = xs.size()  # chunk_size=frames/sub_rate
+        assert cache_size == _cache_size
+        assert chunk_size == -1 or chunk_size == _chunk_size
+        xs = self.embed.out(
+            xs.transpose(1, 2).contiguous().view(b, _chunk_size, channel * f))
```
robin1001, line 276: if chunk_size is -1, the actual chunk is inferred from other dimensions

runtime/core/decoder/torch_asr_decoder.cc
```diff
@@ -104,6 +115,14 @@ DecodeState TorchAsrDecoder::AdvanceDecoding() {
   // If not okay, that means we reach the end of the input
   if (!feature_pipeline_->Read(num_requried_frames, &chunk_feats)) {
     state = DecodeState::kEndFeats;
+    if (num_requried_frames < std::numeric_limits<int>::max() &&
```
robin1001, line 118: Is is for avoiding dynamic slice in Libtorch?

runtime/core/decoder/torch_asr_decoder.h
robin1001: Any way to simplify it? can we use torch::zeros(0) to denote that is an empty cache?

xingchensong:
combine key & value cache will introduce extra concat & slice（12 layers * 2 = 24 extra ops in total），this may hurt performance on other devices such as bpu.

xingchensong:
no，this will introduce extra if-else branch in forward_chunk，which may hurt onnx export

robin1001:
Discussed with @xingchensong , we will refine the following design:
1. combine key_cache and value_cache as attn_cache since they shares the same operation.
2. do not padding the last chunk since 1) padding may result in uncertain result 2) padding result in extra computing cost.
3. it's better to keep next_cache_start in the python code to make simpler interface for caller.

fanlu:
@xingchensong Cool, Great work!
I have some questions, have you checked the rtf significantly increased in the first test round? and the CER value of the new api is a little better than old api, Are there some result of  other test sets?

xingchensong:
> @xingchensong Cool, Great work! I have some questions, have you checked the rtf significantly increased in the first test round? and the CER value of the new api is a little better than old api, Are there some result of other test sets?

Round 1 is a warmup round, we can conclude that APIv2 relies on model warmup heavily. However this is not an issue as long as we have enough warmup data. As for CER, detailed explanations can be found in https://horizonrobotics.feishu.cn/sheets/shtcniLh77AgP6NJAXhd5UHXDwh

kobenaxie:
@xingchensong  How many conformer layers, output_size and hidden_units in the u++conformer? Could you please share the train.cfg of this model ?

xingchensong:
> 

this is a standard u2++ conformer, plz see https://github.com/wenet-e2e/wenet/blob/main/examples/aishell/s0/conf/train_u2%2B%2B_conformer.yaml

xingchensong:
We move this PR to https://github.com/wenet-e2e/wenet/pull/1002, closed.

PR #970, opened 2022-03-07
fix topk bug with small words.txt

yaojunqi:
fix topk bug with small words.txt


yaojunqi (commented):
fix topk bug with small words.txt

robin1001 (approved):

PR #974, opened 2022-03-10
fix invalid variables in tools/format_data.sh

jinsongpan:
"feat-type" and "preprocess-conf" do not take effect in recipes.


robin1001:
parse_options.sh could parse either preprocess-conf or preprocess_conf type parameters, and _ is preferred in coding.

PR #991, opened 2022-03-24
check input.size() > 0 before comparison

eschmidbauer:
- when `input.size()` is `0`, `CHECK_EQ(input.size(), time_stamp.size());` causes crash.
- before comparison, make sure input.size() is not empty

### DIFF
--- a/runtime/core/decoder/torch_asr_decoder.cc
+++ b/runtime/core/decoder/torch_asr_decoder.cc
@@ -222,17 +222,19 @@ void TorchAsrDecoder::UpdateResult(bool finish) {
     if (unit_table_ != nullptr && finish) {
       const std::vector<int>& input = inputs[i];
       const std::vector<int>& time_stamp = times[i];
-      CHECK_EQ(input.size(), time_stamp.size());
-      for (size_t j = 0; j < input.size(); j++) {
-        std::string word = unit_table_->Find(input[j]);
-        int start = j > 0 ? ((time_stamp[j - 1] + time_stamp[j]) / 2 *
-                             frame_shift_in_ms())
-                          : 0;
-        int end = j < input.size() - 1 ? ((time_stamp[j] + time_stamp[j + 1]) /
-                                          2 * frame_shift_in_ms())
-                                       : offset_ * frame_shift_in_ms();
-        WordPiece word_piece(word, offset + start, offset + end);
-        path.word_pieces.emplace_back(word_piece);
+      if (input.size() > 0) {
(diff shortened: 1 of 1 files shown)


robin1001:
You mean CHECK_EQ(0, 0) crash?

eschmidbauer:
> You mean CHECK_EQ(0, 0) crash?

No, I noticed input.size will be 0 and time_stamp.size will be a non-zero positive number and cause crash

robin1001:
Could help to figure out why the time_stamp is not zero?

eschmidbauer:
Sure. I could debug further but this stopped the crash. I will see if I can pinpoint why the values are not equal

eschmidbauer:
closing this, as problem seems to be resolved in latest code

murphypei:
Is there a conclusion to this question?
@robin1001 @eschmidbauer

eschmidbauer:
> Is there a conclusion to this question? I also got this crash @robin1001 @eschmidbauer

i have not been able to reproduce it in the latest code but i suspect it was related to sending "bad" data to the decoder.

murphypei:
> 

Thanks for your reply.

murphypei:
> > Is there a conclusion to this question? I also got this crash @robin1001 @eschmidbauer
> 
> i have not been able to reproduce it in the latest code but i suspect it was related to sending "bad" data to the decoder.

Thanks for your reply.

murphypei:
@robin1001 I got this error even in the latest version of wenet. But the input data is just recording from Android device, I didn't save them and can't reproduce the error, I think it's should be caused by abnormal data. 

So before comparison, I also make sure input.size() is not empty as  @eschmidbauer did.

Mddct:
same problem: 
asr_decoder.cc:143] Check failed: input.size() == time_stamp.size() (0 vs. 9)

robin1001:
> same problem in latest code: asr_decoder.cc:143] Check failed: input.size() == time_stamp.size() (0 vs. 9)

Have you figured it out?

Mddct:
> > same problem in latest code: asr_decoder.cc:143] Check failed: input.size() == time_stamp.size() (0 vs. 9)
> 
> Have you figured it out?

I'm working on it.

Mddct:
When the endpint is triggered, the next chunk we don't have enough feature  for model->forward ， 

and then we have hyp vector<vector> which has empty vector , and times.size() == 0, const std::vector<int>& time_stamp = times[0] is a undefined behavior

Issue #929, opened 2022-01-25
More deletion error when using TLG decode


cliffchen123:
When using TLG to decode, I try to rise acoustic_scale. Although it bring better CER, it also caused more deletion error.
Any good method can adjust weight of acoustic and graphic without causing deletion error?

robin1001:
Adding LM always results in more deletion error.

Arrivederci:
> Adding LM always results in more deletion error.
I meet the same problem, Could you explain why LM always leads to more deletion error?
Thanks

Arrivederci:
@cliffchen123 Did you use the code in wenet runtime? I can't find where to add a del penalty

cliffchen123:
@Arrivederci I add deletion penalty to graph_cost of wfst decoding.
@robin1001 Do you think it is necessary to propose a PR?

Yymax-max:
Adding LM always results in more deletion error.
I also meet the same problem.can you give some idea ,thank

kangj13:
> @Arrivederci I add deletion penalty to graph_cost of wfst decoding. @robin1001 Do you think it is necessary to propose a PR?

I met same problem, can you propose a PR?  Thanks

Arrivederci:
> > @Arrivederci I add deletion penalty to graph_cost of wfst decoding. @robin1001 Do you think it is necessary to propose a PR?
> 
> I met same problem, can you propose a PR? Thanks

@cliffchen123 that would be very helpful

robin1001:
> @Arrivederci I add deletion penalty to graph_cost of wfst decoding. @robin1001 Do you think it is necessary to propose a PR?

please make a PR to show how it works.

PR #994, opened 2022-03-27
[runtime] deletion penalty in wfst decoding

cliffchen123:
solve the issue #929
Adding a deletion penalty reduces deletion errors, but increases substitution and insertion errors.
It also makes decoding slower, so I would use it with less beam or min, max active.
In my test set, setting a deletion penalty of 1 gives better performance.

### DIFF
--- a/runtime/core/decoder/params.h
+++ b/runtime/core/decoder/params.h
@@ -43,6 +43,7 @@ DEFINE_int32(max_active, 7000, "max active states in ctc wfst search");
 DEFINE_int32(min_active, 200, "min active states in ctc wfst search");
 DEFINE_double(beam, 16.0, "beam in ctc wfst search");
 DEFINE_double(lattice_beam, 10.0, "lattice beam in ctc wfst search");
+DEFINE_double(deletion_penalty, 0.0, "deletion penalty in ctc wfst search");
 DEFINE_double(acoustic_scale, 1.0, "acoustic scale for ctc wfst search");
 DEFINE_double(blank_skip_thresh, 1.0,
               "blank skip thresh for ctc wfst search, 1.0 means no skip");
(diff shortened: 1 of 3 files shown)


robin1001:
Any number of your test?

cliffchen123:
I tested six test sets and their deletion errors were all reduced to the same as without LM.
The total number of sentences in the test set is 10,000.

robin1001:
In my opinion, there are two cases in `arc.ilabel == 0`:
1.  LM backoff arc, there is the backoff weight on the arc, typically it's much bigger than 1.0.
2. Other  `arc.ilabel == 0` which is auxiliary for graph building, penalty is not required for this arcs.

Arrivederci:
I did some experiments, found that adding penalty to ctc blank score works better, which increaced fewer insertion error. 
refer to [this paper](https://arxiv.org/pdf/2101.06856.pdf)

yxf0605:
> I tested six test sets and their deletion errors were all reduced to the same as without LM. The total number of sentences in the 6 test sets is 10,000.

How about the wholistic WER changes?

kangj13:
In my experiments, I set deletion_penalty as 1.0 and ctc_blank_penalty as 0.5. 
Using this configuration, the best CER results are gotten.

shunfeichen:
> 
Hi， the best CER results is compared with the results without language model ?

kangj13:
> > 
> 
> Hi， the best CER results is compared with the results without language model ?

with language model

Yymax-max:
> In my experiments, I set deletion_penalty as 1.0 and ctc_blank_penalty as 0.5. Using this configuration, the best CER results are gotten.
where is ctc_blank_penalty?

robin1001:
supported, please see https://github.com/wenet-e2e/wenet/pull/1166

PR #1006, opened 2022-04-02
[doc] refine pretrained model links

robin1001:
We used to provide checkpoint model, runtime x86 model, runtime android model. However, runtime model could be used in both x86 and android, which is tricky for some users.

So we combine them into one quantized model in this PR, which reduces the model size, network traffic, as well as maintainance efforts. And our experiments shows, quantized models show comparable accuracy and better RTF compared with corresponding float model on both x86 and android platforms.


xingchensong (commented):
Is final.pt averaged from multi epoch or it is just the ckpt from the last epoch?
image

robin1001:
it's averaged checkpoint. It's just our convention to name it as `final`

robin1001:
Since we changed the interface, all the runtime models should be re-exported with the new interface. That's the main function of the PR.

xingchensong (approved):

PR #1019, opened 2022-04-07
[runtime] try automake workflow

robin1001:


robin1001:
please see https://github.com/wenet-e2e/wenet/pull/1030.

PR #1023, opened 2022-04-08
[bin] onnx_streaming and onnx_non_streaming

xingchensong:
I find it quite easy to export onnx (either streaming model or non-streaming model) within our recently developed new cache API. The most tricky thing for onnx export is to ensure that all if-else branches in `forward_chunk` will always take the same path with a given decoding setup (i.e., 16/-1, 16/4, 16/0 or -1/-1) (this will turn the dynamic if-else branch into a static branch and `next_cache_start` will be a fixed value). To do this, we can simply adjust the shape of cache & mask tensors for the **first** input chunk according to `chunk_size` and `num_left_chunks`.

# TODO
- [x] check consistency between onnx_model and pytorch_model on random data
- [ ] check CER on aishell-1 testset

# Q&A
1. Why don't we need `jit.script` for if-else branches?
As mentioned above, we force our model to select the same branch for every chunk.

2. Why don't we need `jit.script` for tensor slice?
**From @Mddct ** : ONNX supports dynamic slice for opset>=13, see https://github.com/onnx/onnx/blob/main/docs/Operators.md#Slice

```py
# Author: @Mddct 
import torch
from torch import nn

class BadFirst(nn.Module):
    def __init__(self):
        super().__init__()

    def forward(self, x: torch.Tensor, index: torch.Tensor):

        x_firsts = x[:index]
        return x_firsts


m = BadFirst().eval()
x = torch.ones(44)
index = torch.tensor(0, dtype=torch.int)

res = m(x, index=index) # this works
torch.onnx.export(m, (x, index),  "m.onnx",  opset_version=14)
```

### DIFF
--- a/wenet/bin/export_onnx_cpu.py
+++ b/wenet/bin/export_onnx_cpu.py
@@ -1,0 +1,327 @@
+#!/usr/bin/env python3
+# Copyright (c) 2022, Xingchen Song (sxc19@mails.tsinghua.edu.cn)
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from __future__ import print_function
+
+import argparse
+import os
+import copy
+import sys
+
+import torch
+import yaml
+import numpy as np
+
+from wenet.transformer.asr_model import init_asr_model
+from wenet.utils.checkpoint import load_checkpoint
+from wenet.utils.common import reverse_pad_list, add_sos_eos
+
+try:
+    import onnx
+    import onnxruntime
+except ImportError:
+    print('Please install onnxruntime!')
+    sys.exit(1)
+
+
+def get_args():
+    parser = argparse.ArgumentParser(description='export your script model')
+    parser.add_argument('--config', required=True, help='config file')
+    parser.add_argument('--checkpoint', required=True, help='checkpoint model')
+    parser.add_argument('--output_dir', required=True, help='output directory')
+    parser.add_argument('--chunk_size', required=True,
+                        type=int, help='decoding chunk size')
+    parser.add_argument('--num_decoding_left_chunks', required=True,
+                        type=int, help='cache chunks')
+    parser.add_argument('--beam', required=True,
+                        type=int, help='beam wigth')
+    parser.add_argument('--reverse_weight', default=0.0,
+                        type=float, help='reverse_weight in attention_rescoing')
+    args = parser.parse_args()
+    return args
+
+
+def to_numpy(tensor):
+    if tensor.requires_grad:
+        return tensor.detach().cpu().numpy()
+    else:
+        return tensor.cpu().numpy()
+
+
+def main():
+    torch.manual_seed(777)
+    args = get_args()
+    output_dir = args.output_dir
+    os.system("mkdir -p " + output_dir)
+    os.environ['CUDA_VISIBLE_DEVICES'] = '-1'
+
+    with open(args.config, 'r') as fin:
+        configs = yaml.load(fin, Loader=yaml.FullLoader)
+
+    model = init_asr_model(configs)
+    load_checkpoint(model, args.checkpoint)
+    model.eval()
+    print(model)
+
+    batch = 1
+    chunk_size = args.chunk_size
+    left_chunks = args.num_decoding_left_chunks
+    beam = args.beam
+    reverse_weight = args.reverse_weight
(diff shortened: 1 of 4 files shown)


robin1001:
@xingchensong , please cowork with @lzhin on this PR.

robin1001:
@lzhin, please review and verify the PR.

lzhin:
@robin1001, ok, I'll do it soon.

Mddct:
May we can tun if-else branch code  to tensor operation for export onnx
```python
zeor_tensor = torch.tensor(0)

lt_zero = torch.less(required_cache_size, zeor_tensor)
gt_zero = torch.greater(required_cache_size, zeor_tensor)

tmp_cur_offset = attention_key_size - required_cache_size
tmp_cur_offset = torch.where(torch.greater(tmp_cur_offset, 0), 
                             tmp_cur_offset, 
                             0)

next_cache_size = torch.where(lt_zero, 
                              torch.tensor(0),
                              attention_key_size
                              )

next_cache_size = torch.where(gt_zero,
                              tmp_cur_offset,
                              next_cache_size,
                              )
```

wenet/bin/export_onnx_cpu.py
```diff
@@ -0,0 +1,327 @@
...
+    inputs = (chunk, offset, required_cache_size,
+              att_cache, cnn_cache, att_mask)
+    print("\t\t\033[32mchunk.size(): {}\033[0m\n".format(chunk.size()),
+          "\t\t\033[32moffset: {}\033[0m\n".format(offset),
+          "\t\t\033[32mrequired_cache: {}\033[0m\n".format(required_cache_size),
+          "\t\t\033[32matt_cache.size(): {}\033[0m\n".format(att_cache.size()),
+          "\t\t\033[32mcnn_cache.size(): {}\033[0m\n".format(cnn_cache.size()),
+          "\t\t\033[32matt_mask.size(): {}\033[0m\n".format(att_mask.size()))
+
+    print("\t\033[32mStage-1.2: torch.onnx.export\033[0m")
```
robin1001, line 134: replace print("\t\033[32m{}\033[0m") with a function

robin1001: Can we split the whole processing into functions, seems it is too long now.
I think we can move the verfication code into a new function.

robin1001:
I got another question, how can we save `eos`, `sos`, `subsampling_rage`, `right context`, etc information? what's the best way to do that?

xingchensong:
> I got another question, how can we save `eos`, `sos`, `subsampling_rage`, `right context`, etc information? what's the best way to do that?

from train.yaml and initialized model

https://github.com/wenet-e2e/wenet/pull/1023/files#diff-e4274fd35d26689b022f06cb30cd14be040f8075c97a36521684c02dad9412ecR83-R91

https://github.com/wenet-e2e/wenet/pull/1023/files#diff-e4274fd35d26689b022f06cb30cd14be040f8075c97a36521684c02dad9412ecR278

wenet/bin/export_onnx_cpu.py
xingchensong: done

Mddct:
> I got another question, how can we save `eos`, `sos`, `subsampling_rage`, `right context`, etc information? what's the best way to do that?

can we do this：
torch onnx export four onnx models：
1 encoder
2 decoder
3 ctc decoder
4 meta （just some constant）

and then concat those four onnx graps to one graph：https://github.com/scailable/sclblonnx/blob/master/sclblonnx/merge.py#L213

we can call session.run on a graph using different input names

Mddct:
> > I got another question, how can we save `eos`, `sos`, `subsampling_rage`, `right context`, etc information? what's the best way to do that?
> 
> can we do this： torch onnx export four onnx models： 1 encoder 2 decoder 3 ctc decoder 4 meta （just some constant）
> 
> and then concat those four onnx graps to one graph：https://github.com/scailable/sclblonnx/blob/master/sclblonnx/merge.py#L213
> 
> we can call session.run on a graph using different input names

we can't run session using different input namse, because session is binding to a model with fixed input names.

But i find a keyword in onnx proto: 
Optional Metadata：https://github.com/onnx/onnx/blob/main/docs/IR.md
Maybe we can keep sos ... here

xingchensong:
we can write and read metadatas like this:

```py
# write
onnx_encoder = onnx.load(encoder_outpath)
for (k, v) in args.items():
    meta = onnx_encoder.metadata_props.add()
    meta.key, meta.value = str(k), str(v)
onnx.save(onnx_encoder, encoder_outpath)

# read
ort_session = onnxruntime.InferenceSession(encoder_outpath)
meta = ort_session.get_modelmeta()
print("\t\tcustom_metadata_map={}".format(meta.custom_metadata_map))
```

Mddct:
onnxruntime support openvino provider。Do we have any plan to support openvino？

robin1001:
> onnxruntime support openvino provider。Do we have any plan to support openvino？

openvino is not in our paln. The user could switch to openvino backend if they do want it.

robin1001 (approved):

lzhin:
> onnxruntime support openvino provider。Do we have any plan to support openvino？

hello, could you help me look at this [problem](https://github.com/microsoft/onnxruntime/issues/11257)?

PR #1029, opened 2022-04-11
binding runtime for python and other language

Mddct:
features：
- python binding
-  go binding
- c interface
- label check binding

TODO:
- [ ] fix streaming decoding hang


robin1001:
Great job, I will go through the details soon. @xingchensong @pengzhendong please follow the PR if you have time.

Mddct:
> WoW~ 
> Not sure if it is possible to have a binding for decoder only? 
> So that we may replace the pytorch ctc: https://github.com/wenet-e2e/wenet/blob/main/wenet/transformer/asr_model.py#L330 

The decoder at runtime has many states: model, feature.pipeline, etc.

  We can rewrite a simple decoder only multiplexing ctc prefix beam search, which has chunk state

Mddct:
closed due to new api

PR #1049, opened 2022-04-20
Support WeNet to run on Graphcore IPU

biofoolgreen:
First commit to support Graphcore IPU.


requirements.txt
```diff
@@ -14,3 +14,4 @@ flake8-pyi==20.5.0
 mccabe
 pycodestyle==2.6.0
 pyflakes==2.2.0
+https://download.pytorch.org/whl/cpu/torchaudio-0.10.0%2Bcpu-cp36-cp36m-linux_x86_64.whl
```
robin1001, line 17: torchaudio is installed by conda, so it is not required here.

wenet/bin/recognize.py
```diff
@@ -128,10 +128,10 @@ def main():
     symbol_table = read_symbol_table(args.dict)
     test_conf = copy.deepcopy(configs['dataset_conf'])
 
-    test_conf['filter_conf']['max_length'] = 102400
-    test_conf['filter_conf']['min_length'] = 0
-    test_conf['filter_conf']['token_max_length'] = 102400
-    test_conf['filter_conf']['token_min_length'] = 0
+    test_conf['filter_conf']['max_length'] = 1200
```
robin1001, line 131: why change the default parameter here? is there any limits on IPU?

wenet/bin/train.py
```diff
@@ -243,7 +282,11 @@ def main():
         device = torch.device('cuda' if use_cuda else 'cpu')
         model = model.to(device)
 
-    optimizer = optim.Adam(model.parameters(), **configs['optim_conf'])
+    optimizer = poptorch.optim.Adam(
+        model.parameters(),
+        accum_type=torch.float16,
```
robin1001, line 287: condational set it  to torch.float16?

wenet/dataset/dataset.py
```diff
@@ -187,3 +187,31 @@ def Dataset(data_type,
...
+        self.max_target_length = max_target_length
+        assert type in ['float32', 'float16']
+        self.type = type
+
+    def __call__(self, batch):
+        batch_size = len(batch)
+        feature_length = torch.cat([i[3] for i in batch])
+        target_length = torch.cat([i[4] for i in batch])
+        padded_feature = torch.zeros(
+            size=[batch_size, self.max_feature_length, batch[0][1].size(2)])
```
robin1001, line 208: What's the difference of this code compared to the pading function in processor.py?

wenet/transformer/asr_model.py
```diff
@@ -718,6 +727,20 @@ def init_asr_model(configs):
         decoder = BiTransformerDecoder(vocab_size, encoder.output_size(),
                                        **configs['decoder_conf'])
     ctc = CTC(vocab_size, encoder.output_size())
+
```
robin1001, line 730: If it is caused by torch.jit.export? we can simply remove jit check in train.py.

wenet/transformer/convolution.py
```diff
@@ -66,7 +66,7 @@ def __init__(self,
             self.norm = nn.BatchNorm1d(channels)
         else:
             self.use_layer_norm = True
-            self.norm = nn.LayerNorm(channels)
+            self.norm = nn.LayerNorm(channels, eps=1e-3)
```
robin1001, line 69: any explanation or comment about setting it to 1e-3?

wenet/transformer/label_smoothing_loss.py
```diff
@@ -48,11 +48,12 @@ def __init__(self,
                  normalize_length: bool = False):
         """Construct an LabelSmoothingLoss object."""
         super(LabelSmoothingLoss, self).__init__()
-        self.criterion = nn.KLDivLoss(reduction="none")
+        self.size = size
```
robin1001, line 51: KLD is not support on IPU?

tongjinle123: we have rewritten a label smoothing loss, and they are tested to be equal,
the original code's scatter op have some problem in performance.

wenet/transformer/convolution.py
tongjinle123: our model is mainly trained on fp16, and the precision is set to 1e-3 to prevent numerical problem

wenet/transformer/asr_model.py
tongjinle123: yes, I make such walk around rather than just remove the jit because I found those exported function is used in runtime/, 
could be removed if it is ok.

wenet/dataset/dataset.py
tongjinle123: the padding shape is statics while the original function is dynamic

wenet/bin/recognize.py
tongjinle123: is reset to default, a wrong change in the very beginning

requirements.txt
tongjinle123: our SDK environment is using python3.6, and the installation steps may have to follow the IPU_README.md.

there is also a problem for the checking below: `UnitTest / unit-test (ubuntu-latest, 3.8, 1.10.0, 0.10.0) (pull_request) Failing after 2m — unit-test (ubuntu-latest, 3.8, 1.10.0, 0.10.0)`

wenet/bin/train.py
tongjinle123: modified to follow the `precision`

PR #1051, opened 2022-04-22
fix check CHECK_EQ(input.size(), time_stamp.size())

Mddct:
related pr: https://github.com/wenet-e2e/wenet/pull/991


Mddct:
have a better solution

PR #1053, opened 2022-04-22
runtime cpu onnx

lzhin:
draft pr

### DIFF
--- a/runtime/core/decoder/onnx_asr_model.cc
+++ b/runtime/core/decoder/onnx_asr_model.cc
@@ -1,7 +1,7 @@
 // Copyright 2020 Mobvoi Inc. All Rights Reserved.
 // Author: binbinzhang@mobvoi.com (Binbin Zhang)
 //         di.wu@mobvoi.com (Di Wu)
-// Copyright 2022 Mobvoi Inc. All Rights Reserved.
+// Copyright 2022 Huya Inc. All Rights Reserved.
 // Author: lizexuan@huya.com
 
 #include "decoder/onnx_asr_model.h"
@@ -34,7 +34,6 @@ void OnnxAsrModel::Read(const std::string& model_dir, const int num_threads) {
 
     Ort::Session ctc_session{env_, ctc_onnx_path.data(), session_options};
     ctc_session_ = std::make_shared<Ort::Session>(std::move(ctc_session));
-
   } catch (std::exception const& e) {
     LOG(FATAL) << "error when load onnx model";
     exit(0);
@@ -149,10 +148,8 @@ void OnnxAsrModel::Reset() {
         memory_info_, att_cache_.data(), att_cache_.size(), att_cache_shape, 4);
 
     att_mask_.resize(0);
-    const int64_t att_mask_shape[] = {0, 0, 0};
-    att_mask_ort_ = Ort::Value::CreateTensor<bool>(
-        memory_info_, reinterpret_cast<bool*>(att_mask_.data()),
-        att_mask_.size(), att_mask_shape, 3);
+    input_names_.erase(input_names_.begin() + 2);
+    input_names_.pop_back();
   }
 
   cnn_cache_.resize(num_blocks_ * encoder_output_size_ *
@@ -183,6 +180,16 @@ void OnnxAsrModel::ForwardEncoderFunc(
       model_input.emplace_back(chunk_feats[i][j]);
     }
   }
+  int num_fill =
+      att_mask_.size() + 1 + right_context_ - subsampling_rate_ - num_frames;
+  if (num_fill > 0) {
+    num_frames += num_fill;
+    for (size_t i = 0; i < num_fill; ++i) {
+      for (int j = 0; j < feature_dim; j++) {
+        model_input.emplace_back(0);
+      }
+    }
+  }
   const int64_t input_shape[3] = {1, num_frames, feature_dim};
   Ort::Value input_tensor = Ort::Value::CreateTensor<float>(
       memory_info_, model_input.data(), model_input.size(), input_shape, 3);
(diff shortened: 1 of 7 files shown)


runtime/core/decoder/onnx_asr_model.cc
```diff
@@ -0,0 +1,347 @@
...
+  Ort::Value requried_cache_size_tensor = Ort::Value::CreateTensor<int64_t>(
+      memory_info_, &requried_cache_size, 1, std::vector<int64_t>{}.data(), 0);
+
+  std::vector<Ort::Value> tensors;
+  tensors.emplace_back(std::move(input_tensor));
+  tensors.emplace_back(std::move(offset_tensor));
+  tensors.emplace_back(std::move(requried_cache_size_tensor));
+  tensors.emplace_back(std::move(att_cache_ort_));
+  tensors.emplace_back(std::move(cnn_cache_ort_));
+  tensors.emplace_back(std::move(att_mask_ort_));
```
xingchensong, line 201: In 16/4 mode, `at_mask_ort_` should be reused for every chunk, however, `move` will make `att_mask_ort_` inaccessible, leading to segment fault. We should recreate `att_mask_ort_` everytime we feed a new chunk.

xingchensong: option-2: there is no need to maintain `att_mask_ort_` as a class member, just construct `att_mask_ort` on-the-fly wherever needed.

runtime/core/decoder/onnx_asr_model.h
```diff
@@ -0,0 +1,73 @@
...
+      Ort::MemoryInfo::CreateCpu(OrtDeviceAllocator, OrtMemTypeCPU);
+  std::vector<char*> input_names_ = {
+      "chunk",     "offset",    "required_cache_size",
+      "att_cache", "cnn_cache", "att_mask"};
+  const char* output_names_[3] = {"output", "r_att_cache", "r_cnn_cache"};
+  const char* decode_input_names_[3] = {"hyps", "hyps_lens", "encoder_out"};
+  const char* decode_output_names_[2] = {"score", "r_score"};
+
+  const char* ctc_input_names_[1] = {"hidden"};
+  const char* ctc_output_names_[1] = {"probs"};
```
xingchensong, line 60: we can dynamically read node names like this: 

```cpp
void OnnxAsrModel::GetInputOutputInfo(
    const std::shared_ptr<Ort::Session>& session,
    std::vector<const char*>* in_names,
    std::vector<const char*>* out_names) {
  Ort::AllocatorWithDefaultOptions allocator;
  // Input info
  int num_nodes = session->GetInputCount();
  in_names->resize(num_nodes);
  for (int i = 0; i < num_nodes; ++i) {
    char* name = session->GetInputName(i, allocator);
    Ort::TypeInfo type_info = session->GetInputTypeInfo(i);
    auto tensor_info = type_info.GetTensorTypeAndShapeInfo();
    ONNXTensorElementDataType type = tensor_info.GetElementType();
    std::vector<int64_t> node_dims = tensor_info.GetShape();
    std::stringstream shape;
    for (auto j : node_dims) {shape << j; shape << " ";}
    VLOG(2) << "\tInput " << i << " : name=" << name << " type=" << type
            << " dims=" << shape.str();
    (*in_names)[i] = name;
  }
  // Output info
  num_nodes = session->GetOutputCount();
  out_names->resize(num_nodes);
  for (int i = 0; i < num_nodes; ++i) {
    char* name = session->GetOutputName(i, allocator);
    Ort::TypeInfo type_info = session->GetOutputTypeInfo(i);
    auto tensor_info = type_info.GetTensorTypeAndShapeInfo();
    ONNXTensorElementDataType type = tensor_info.GetElementType();
    std::vector<int64_t> node_dims = tensor_info.GetShape();
    std::stringstream shape;
    for (auto j : node_dims) {shape << j; shape << " ";}
    VLOG(2) << "\tOutput " << i << " : name=" << name << " type=" << type
            << " dims=" << shape.str();
    (*out_names)[i] = name;
  }
}
```

runtime/core/decoder/onnx_asr_model.cc
```diff
@@ -0,0 +1,362 @@
...
+  } else {
+    const int64_t att_mask_shape[] = {1, 1, att_mask_.size()};
+    att_mask_ort_ = Ort::Value::CreateTensor<bool>(
+        memory_info_, reinterpret_cast<bool*>(att_mask_.data()),
+        att_mask_.size(), att_mask_shape, 3);
+    tensors.emplace_back(std::move(requried_cache_size_tensor));
+    tensors.emplace_back(std::move(att_cache_ort_));
+    tensors.emplace_back(std::move(cnn_cache_ort_));
+    tensors.emplace_back(std::move(att_mask_ort_));
+  }
```
xingchensong, line 217: emplace_back ort_tensors according to input node names, in this way there is no need to manually control what should be emplaced:

```cpp
std::vector<Ort::Value> inputs;
  for (auto name : encoder_in_names_) {
    if (!strcmp(name, "chunk")) {
      inputs.emplace_back(std::move(feats_ort));
    } else if (!strcmp(name, "offset")) {
      inputs.emplace_back(std::move(offset_ort));
    } else if (!strcmp(name, "required_cache_size")) {
      inputs.emplace_back(std::move(required_cache_size_ort));
    } else if (!strcmp(name, "att_cache")) {
      inputs.emplace_back(std::move(att_cache_ort_));
    } else if (!strcmp(name, "cnn_cache")) {
      inputs.emplace_back(std::move(cnn_cache_ort_));
    } else if (!strcmp(name, "att_mask")) {
      inputs.emplace_back(std::move(att_mask_ort));
    }
  }
```

runtime/core/decoder/onnx_asr_model.h
```diff
@@ -0,0 +1,73 @@
...
+
+  Ort::Value att_cache_ort_{nullptr};
+  Ort::Value cnn_cache_ort_{nullptr};
+  Ort::Value att_mask_ort_{nullptr};
+  std::vector<Ort::Value> encoder_outs_;
+  std::vector<float> att_cache_;
+  std::vector<float> cnn_cache_;
+  std::vector<uint8_t> att_mask_;
```
xingchensong, line 68: similar to  torch_asr_model, we only need to main `att_cache_ort_` `cnn_cache_ort_` `encoder_outs_` as class members, other variables can be generated on-the-fly.

runtime/core/decoder/onnx_asr_model.cc
xingchensong: since we have so many decoding cases, i.e.,  streaming model with 16/-1 16/4/ -1/-1 16/0, non-streaming model with -1/-1, different case has different in/out nodes, it is hard to manually decide which node(tensor) should be emplaced.

runtime/core/decoder/onnx_asr_model.cc
```diff
@@ -0,0 +1,362 @@
...
+  int num_fill =
+      att_mask_.size() + 1 + right_context_ - subsampling_rate_ - num_frames;
+  if (num_fill > 0) {
+    num_frames += num_fill;
+    for (size_t i = 0; i < num_fill; ++i) {
+      for (int j = 0; j < feature_dim; j++) {
+        model_input.emplace_back(0);
+      }
+    }
+  }
```
xingchensong, line 192: I find a easy way to handle the last chunk. Just add a slice op in attention.py https://github.com/wenet-e2e/wenet/blob/main/wenet/transformer/attention.py#L90-L94

before:
```py
if mask.size(2) > 0 :  # time2 > 0
            mask = mask.unsqueeze(1).eq(0)  # (batch, 1, *, time2)
            scores = scores.masked_fill(mask, -float('inf'))
            attn = torch.softmax(scores, dim=-1).masked_fill(
                mask, 0.0)  # (batch, head, time1, time2)
```

after:
```py
if mask.size(2) > 0 :  # time2 > 0
            mask = mask.unsqueeze(1).eq(0)  # (batch, 1, *, time2)
            mask = mask[:, :, :, :scores.size(-1)]
            scores = scores.masked_fill(mask, -float('inf'))
            attn = torch.softmax(scores, dim=-1).masked_fill(
                mask, 0.0)  # (batch, head, time1, time2)
```

runtime/core/decoder/onnx_asr_model.h
```diff
@@ -0,0 +1,73 @@
...
+  float ComputeAttentionScore(const float* prob, const std::vector<int>& hyp,
+                              int eos, int decode_out_len);
+
+ private:
+  int encoder_output_size_ = 0;
+  int num_blocks_ = 0;
+  int cnn_module_kernel_ = 0;
+  int head_ = 0;
+
+  Ort::Env env_;
```
xingchensong, line 45: Maybe env can be moved to function `OnnxAsrModel::Read() `

runtime/core/decoder/onnx_asr_model.cc
xingchensong: when export onnxmodel, we can keep dynamic axis for `chunk` to allow different input shapes for chunk tensors

xingchensong:
Closed, we refine our code in  another PR #1103

PR #1056, opened 2022-04-23
[tools] normalize the audio before resample

Mashiro009:
According to [pytorchaudio issues](https://github.com/pytorch/audio/issues/2294), resample can't process the int audio.
And `waveforms, sample_rate = sox.load(wav, normalize=False)`, so the type of waveforms from wav is int16.
Therefore normalize the audio before resample by dividing (1<<15).


robin1001 (approved):

robin1001:
Good Catch!

PR #1059, opened 2022-04-25
[runtime] add wenet runtime api

robin1001:
We design the wenet api for other language bingding in the future, and we refer https://github.com/alphacep/vosk-api/blob/master/src/vosk_api.h for the design.

I will add the python binding after this PR, @Mddct , I think you can add the go bingding in the future.


robin1001:
@Mddct , please review the PR.

pengzhendong (approved):

xingchensong (approved):

PR #1066, opened 2022-04-28
[binding] add python binding

robin1001:


robin1001:
@Mddct , python binding, please follow the PR.

runtime/binding/python/py/decoder.py
```diff
@@ -0,0 +1,40 @@
...
+    def decode(self, pcm: bytes, last: bool = True) -> str:
+        """ Decode the input data
+
+        Args:
+            pcm: wav pcm
+            last: if it is the last package of the data
+        """
+        assert isinstance(pcm, bytes)
+        finish = 1 if last else 0
+        _wenet.wenet_decode(self.d, pcm, len(pcm), finish)
```
Mddct, line 38: C API asserts can cause code to crash when the data is not in PCM format.
Can we add a friendly try-catch-exception

runtime/binding/python/py/decode_main.py
```diff
@@ -0,0 +1,26 @@
+import sys
```
Mddct, line 1: Movedecoder_main.py to parend direcnionary？
It‘s looks like an example

Mddct (commented):
Done

runtime/binding/python/py/decoder.py
robin1001: Maybe refine it in the future pr.

runtime/binding/python/py/decode_main.py
robin1001: it's a temp test, will refine it in the future.

PR #1072, opened 2022-04-30
[.github] add python wheels workflow

robin1001:
Finally it works, I will refine it in the future.


runtime/binding/python/CMakeLists.txt
```diff
@@ -22,17 +22,40 @@ if(NOT MSVC)
...
+endif()
+FetchContent_Declare(libtorch
+  URL      ${LIBTORCH_URL}
+  URL_HASH ${URL_HASH}
 )
-list(APPEND CMAKE_PREFIX_PATH "${TORCH_DIR}")
-find_package(Torch REQUIRED PATHS ${TORCH_DIR} NO_DEFAULT_PATH)
-set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} ${TORCH_CXX_FLAGS}")
-message(STATUS ${TORCH_CXX_FLAGS})
-message(STATUS "${CMAKE_CXX_FLAGS}")
```
xingchensong, line 35: Why rollback to libtorch？

robin1001: The feature is still under developing, i'm trying to figure out the right way to make it work.

PR #1085, opened 2022-05-06
[runtime] avoid setting the num_interop_threads repeatedly

pengzhendong:


robin1001:
We will use multiple backends(libtorch/onnx/j3/...) in the future,  any better abstraction for this?

robin1001:
How about add one static function like `InitEngineThreads` in AsrModel, so the calling is bind to the backend engine we used.

robin1001 (approved):

PR #1090, opened 2022-05-07
[runtime] refactoring unittest

pengzhendong:


robin1001 (changes requested):
please move test dir to core since all of source code are in core.

robin1001 (approved):

PR #1092, opened 2022-05-08
[binding] use json format

robin1001:


pengzhendong:
Why not use the json tool from boost?

robin1001:
> Why not use the json tool from boost?

boost is too heavy. Here we want to keep it simple and small.

pengzhendong (approved):

PR #1107, opened 2022-05-14
fix word alignment timestamp

xiangxyq:
I have some confusions in my pr, time_stamp_gap_ is not a well variable to define its value, or just use frame_shift_in_ms(), maybe time_stamp_gap_ move to param.h that can be configured.

here just use a gap to limit word align timestamp more accurate, I think if combine blank timestamp, the result will better, but that seems too hard.


robin1001 (approved):

robin1001:
It's okay for the refinement.

PR #1112, opened 2022-05-14
[runtime/onnx] fix the bug of OrtEnv singleton lifetime.

Duum:
The OrtEnv singleton will be release, When the Env object destruct. So need use static mark Env live in whole process.


xingchensong:
Why should we keep `Env` alive? It seems `Env` is only used when constructing sessions.

PR #1131, opened 2022-05-18
[runtime] decoder_main supports multithreading

lzhin:


robin1001:
It's better if we use a thread pool here. How about this https://github.com/progschj/ThreadPool?

lzhin:
> 

ok, I'll revise it later.

robin1001:
okay, not urgent, just keep looking at it and make it best.

lzhin:
> There's a simple version of thread pool in tfseeving https://github.com/tensorflow/serving/blob/2b20315418346837a26a11f640fed0da90048a08/tensorflow_serving/util/net_http/internal/fixed_thread_pool.h

thanks

runtime/core/bin/decoder_main.cc
```diff
@@ -14,60 +15,39 @@
...
 DEFINE_string(wav_path, "", "single wave path");
 DEFINE_string(wav_scp, "", "input wav scp");
 DEFINE_string(result, "", "result output file");
 DEFINE_bool(continuous_decoding, false, "continuous decoding mode");
+DEFINE_int32(thread_num, 1, "num of decode thread");
 
-int main(int argc, char *argv[]) {
-  gflags::ParseCommandLineFlags(&argc, &argv, false);
-  google::InitGoogleLogging(argv[0]);
+std::shared_ptr<wenet::DecodeOptions> decode_config_;
```
robin1001, line 28: do not use postfix _, for global variable,  prefix `g_`  is suggested.

robin1001, line 28: do not use postfix _, for global variable,  prefix `g_`  is suggested.

runtime/core/bin/decoder_main.cc
```diff
@@ -129,13 +111,51 @@ int main(int argc, char *argv[]) {
...
+    result_.open(FLAGS_result, std::ios::out);
+  }
+
+  ThreadPool pool(FLAGS_thread_num);
+  for (auto &wav : waves) {
+    pool.enqueue(decode, wav);
+  }
+
+  while (has_done_ != waves.size()) {}
+
```
robin1001, line 155: You can just use the following method to shutdown the thread pool.
``` c++
{
    ThreadPool pool(FLAGS_thread_num);
    for (auto &wav : waves) {
      pool.enqueue(decode, wav);
    }
}
```

Sometimes some decoding fails, then the program will be blocked forever in 

``` c++
while (has_done_ != waves.size()) {}
```

robin1001 (commented):
please see inline.

runtime/core/bin/decoder_main.cc
```diff
@@ -5,73 +5,47 @@
...
-  std::ofstream result;
-  if (!FLAGS_result.empty()) {
-    result.open(FLAGS_result, std::ios::out);
-  }
-  std::ostream &buffer = FLAGS_result.empty() ? std::cout : result;
-
-  int total_waves_dur = 0;
-  int total_decode_time = 0;
-  for (auto &wav : waves) {
+void * decode(std::pair<std::string, std::string> wav) {
```
xingchensong, line 33: `void* decode`

see .clang-format [link](https://github.com/wenet-e2e/wenet/blob/main/.clang-format#L77)
```.clang-format
PointerAlignment: Left
```

robin1001 (approved):

PR #1147, opened 2022-05-21
[bin] hard code nbest of exporting onnx

pengzhendong:


robin1001:
`nbest` is not required for onnx exporting, so just remove it.

robin1001:
`nbest` is not required for onnx exporting, so just remove it.

xingchensong:
> `nbest` is not required for onnx exporting, so just remove it.

Agreed. This is a legacy issue, in the first version of `export_onnx_cpu`,  `beam`(or `nbest`) was made static.

xingchensong (approved):

PR #1149, opened 2022-05-22
[runtime] copy onnx dlls to cmake_binary_dir

pengzhendong:
Please see:
* https://github.com/microsoft/onnxruntime/issues/11480
* https://github.com/microsoft/onnxruntime/issues/6243

### DIFF
--- a/runtime/core/cmake/onnx.cmake
+++ b/runtime/core/cmake/onnx.cmake
@@ -17,5 +17,10 @@ FetchContent_Declare(onnxruntime
   URL_HASH ${URL_HASH}
 )
 FetchContent_MakeAvailable(onnxruntime)
-include_directories("${onnxruntime_SOURCE_DIR}/include")
-link_directories("${onnxruntime_SOURCE_DIR}/lib")
+include_directories(${onnxruntime_SOURCE_DIR}/include)
+link_directories(${onnxruntime_SOURCE_DIR}/lib)
+
+if(MSVC)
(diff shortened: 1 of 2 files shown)


runtime/core/cmake/onnx.cmake
```diff
@@ -17,5 +17,10 @@ FetchContent_Declare(onnxruntime
   URL_HASH ${URL_HASH}
 )
 FetchContent_MakeAvailable(onnxruntime)
-include_directories("${onnxruntime_SOURCE_DIR}/include")
-link_directories("${onnxruntime_SOURCE_DIR}/lib")
+include_directories(${onnxruntime_SOURCE_DIR}/include)
+link_directories(${onnxruntime_SOURCE_DIR}/lib)
+
+if(MSVC)
```
robin1001, line 23: why we should copy the dll for windows?

pengzhendong: The same as librotch: https://github.com/wenet-e2e/wenet/blob/167594edd69b29614e234e5c9221a01b76c1ee3d/runtime/core/cmake/libtorch.cmake#L33-L36

> Seems like load preferences aren't preferring a local copy of onnxruntime.dll over the one built into Windows

robin1001: I see, have you tried the windows binding, the windows binding size is much smaller than linux/macos. There may be some problem.

robin1001 (approved):

PR #1154, opened 2022-05-22
[model] fix unused parameters in `encoder_layer.py` and `decoder_layer.py`

zyzisyz:
Hi,

I find some unused parameters at [`wenet/wenet/transformer/encoder_layer.py`](https://github.com/wenet-e2e/wenet/blob/6f884838c460510d8f6943ce1a3bc908ee40d349/wenet/transformer/encoder_layer.py#L55) and [`wenet/wenet/transformer/decoder_layer.py`](https://github.com/wenet-e2e/wenet/blob/6f884838c460510d8f6943ce1a3bc908ee40d349/wenet/transformer/decoder_layer.py#L55).

The default configuration of `concat_after` is `False`. 
This will lead the `self.concat_linear` to be unsed and bring redundancy in whole Conformer network.

Therefore, I add if statements to fix this problem.

### DIFF
--- a/wenet/transformer/decoder_layer.py
+++ b/wenet/transformer/decoder_layer.py
@@ -52,8 +52,9 @@ class DecoderLayer(nn.Module):
         self.dropout = nn.Dropout(dropout_rate)
         self.normalize_before = normalize_before
         self.concat_after = concat_after
-        self.concat_linear1 = nn.Linear(size + size, size)
-        self.concat_linear2 = nn.Linear(size + size, size)
+        if self.concat_after == True:
+            self.concat_linear1 = nn.Linear(size + size, size)
+            self.concat_linear2 = nn.Linear(size + size, size)
 
     def forward(
         self,
--- a/wenet/transformer/decoder_layer.py
+++ b/wenet/transformer/decoder_layer.py
@@ -55,6 +55,9 @@ class DecoderLayer(nn.Module):
         if self.concat_after:
             self.concat_linear1 = nn.Linear(size + size, size)
             self.concat_linear2 = nn.Linear(size + size, size)
+        else:
+            self.concat_linear1 = nn.Identity()
+            self.concat_linear2 = nn.Identity()
 
     def forward(
         self,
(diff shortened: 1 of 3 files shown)


zyzisyz:
@xingchensong

xingchensong (commented):
In `ConformerEncoderLayer`, apply same refactoring with `TransformerEncoderLayer`

xingchensong:
PLZ try to verify whether we can re-export models from the pre-trained [checkpoint](https://github.com/wenet-e2e/wenet/blob/main/docs/pretrained_models.md). Since the pre-trained ckpt saved the params of concat_linear and [load_state_dict](https://github.com/wenet-e2e/wenet/blob/main/wenet/utils/checkpoint.py#L20) requires strict matches between ckpt and model definition, I'm a bit worried about whether we can do this if we remove `concat_linear`

robin1001:
Have you tested the code? Does the change pass the JIT checking in training?

xingchensong:
Here are two solutions if we cannot do such a re-export:

Solution-1: 
```py
model.load_state_dict(checkpoint) --> model.load_state_dict(checkpoint, strict=False)
``` 

Solution-2:
add some comments to tell users how to do it successfully.

robin1001:
I prefer the 1st solution.

xingchensong:
If JIT checking cannot be passed, we can try:

```py
if self.concat_after:
    self.concat_linear1 = nn.Linear(size + size, size)
    self.concat_linear2 = nn.Linear(size + size, size)
else:
    self.concat_linear1 = nn.Identity()
    self.concat_linear2 = nn.Identity()
```

robin1001:
Yes, I was thinking the same solution as yours.

zyzisyz:
I use pytorch checkpoint from aishell to test JIT export.

It works if we make all above change.

```Python
# wenet/wenet/utils/checkpoint.py
model.load_state_dict(checkpoint) --> model.load_state_dict(checkpoint, strict=False)

# wenet/wenet/transformer/decoder_layer.py
if self.concat_after:
    self.concat_linear1 = nn.Linear(size + size, size)
    self.concat_linear2 = nn.Linear(size + size, size)
else:
    self.concat_linear1 = nn.Identity()
    self.concat_linear2 = nn.Identity()
```

xingchensong (approved):
good catch

PR #1166, opened 2022-05-26
add length penalty

boji123:
In TLG equipped decoder called 'ctc_wfst_beam_search', the del/ins error ratio is much higher than the one without it called 'ctc_prefix_beam_search'

This PR provides a solution to add length penalty in TLG fst path. Penalty only add on the node where a new word occurs. Besides, it also considered the penalty when caculating beam cut weight.

I also notice another potential solution  to add delete penalty in process noemitting function, but it is added on every new node(i lable=0). It may cause some misunderstanding because a word may last for a long time. See [delete penalty](https://github.com/wenet-e2e/wenet/pull/994)

By comparision, my solution performs better than it. I realize and test both solutions, tables below are the comparison result.

These experiments are performed on aishell-1 using the default decoder.sh and the provided U2++ Conformer pretrained model. Notice that I use the latest chunk cache operators thus the result performs a little better than the number in S0.

Result with parameter ctc weight 0.3, reverse weight 0.5, chunksize -1
image
image

Result with parameter ctc weight 0.3, reverse weight 0.5, chunksize 16
image
image

In addition, my solution shows stable improvement in my application scenario (Esports live streaming).

Any other experiments on open source data is needed, please commet.

### DIFF
--- a/runtime/core/decoder/params.h
+++ b/runtime/core/decoder/params.h
@@ -51,6 +51,9 @@ DEFINE_double(lattice_beam, 10.0, "lattice beam in ctc wfst search");
 DEFINE_double(acoustic_scale, 1.0, "acoustic scale for ctc wfst search");
 DEFINE_double(blank_skip_thresh, 1.0,
               "blank skip thresh for ctc wfst search, 1.0 means no skip");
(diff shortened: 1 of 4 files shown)


boji123:
@robin1001 
Addition experienment about acoustic scale
image

robin1001 (approved):

PR #1178, opened 2022-05-30
[runtime] wenet_api support onnx

pengzhendong:


runtime/binding/python/CMakeLists.txt
```diff
@@ -4,9 +4,11 @@ project(wenet VERSION 0.1)
 
 option(CXX11_ABI "whether to use CXX11_ABI libtorch" OFF)
 option(FST_HAVE_BIN "whether to build fst binaries" OFF)
-set(ONNX OFF)  # ONNX is not used
 set(CMAKE_VERBOSE_MAKEFILE OFF)
 
+option(TORCH "whether to build with Torch" OFF)
```
robin1001, line 9: use set instead of option since we only support TORCH in our plan.
```
set(TORCH ON)
set(ONNX OFF)
```

robin1001: we can just build the api and api_main when `TORCH` is enable in `runtime/x86`

robin1001:
please see https://github.com/wenet-e2e/wenet/pull/1195

PR #1189, opened 2022-06-01
[doc] add roadmap

robin1001:


ROADMAP.md
```diff
@@ -0,0 +1,41 @@
...
+- [ ] Self training, streaming
+- [ ] Light weight, low latency, on-device model exploration
+- [ ] Audio-Visual speech recognition
+- [ ] Platforms
+  - [ ] Rasphberry Pi
+  - [ ] Harmony OS
+- [ ] ASIC XPU
+  - [ ] Horizon Journey
+  - [ ] GraphCore
+  - [ ] TO ADD
```
xingchensong, line 26: 昆仑芯？

robin1001: It depends, I think there will be more and more XPUs for WeNet.  We can add it one by one in the future if they are intergrated in WeNet.

PR #1194, opened 2022-06-03
[runtime] cmake module for websocket/bin,  and fix dependency

robin1001:


runtime/core/websocket/CMakeLists.txt
```diff
@@ -0,0 +1,5 @@
+add_library(websocket STATIC
+  websocket_client.cc
+  websocket_server.cc
+)
+target_link_libraries(websocket PUBLIC decoder)
```
pengzhendong, line 5: Is it necessary to build the server and client together? Because the clinet does not need the decoder library.

robin1001: It's just for unified cmake for every module.

robin1001: Just like what  we did  for `grpc` before.

pengzhendong (approved):

PR #1205, opened 2022-06-08
debug cache if

boji123:
watch this version as reference
https://github.com/wenet-e2e/wenet/blob/fd22df3a799568bf178258f6b1f576929e9bb43b/runtime/core/decoder/torch_asr_decoder.cc#L115


boji123:
will cause error when chunksize = 1

robin1001 (approved):

PR #1211, opened 2022-06-10
[utils] fix:pretrained encoder cmvn

pgzhang:
when load pretrained encoder, the global cmvn mean and istd value will changed to the cmvn of the datasets for pretrained model, but it should be the cmvn of the current datasets

### DIFF
--- a/wenet/utils/checkpoint.py
+++ b/wenet/utils/checkpoint.py
@@ -93,7 +93,8 @@ def load_trained_modules(model: torch.nn.Module, args: None):
         partial_state_dict = OrderedDict()
         for key, value in model_state_dict.items():
             if any(key.startswith(m) for m in modules):
-                partial_state_dict[key] = value
+                if 'global_cmvn' not in key:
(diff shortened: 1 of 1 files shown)


Emiyassstar:
I think it would be better to use  cmvn of pretrained model，if you changed cmvn , the pretraining weights will not match

pgzhang:
> 

I conducted an experiment, and find that using custom data cmvn have a little improvement in wer(0.4~1%), compared to using pretrained model cmvn. It seems that using cmvn of custom or pretrained data have very little impact on the model

Emiyassstar:
> > 
> 
> I conducted an experiment, and find that using custom data cmvn have a little improvement in wer(0.4~1%), compared to using pretrained model cmvn. It seems that using cmvn of custom or pretrained data have very little impact on the model

if you want use new cmvn , you don't need change the code , just set --enc_init_mods like 'encoder.embed,encoder.encoders.0,encoder.encoders.1...' ，this will not use the previous cmvn.

PR #1219, opened 2022-06-11
[binding]: add decode opts

Mddct:


runtime/core/api/wenet_api.cc
```diff
@@ -92,6 +92,14 @@ class Recognizer {
     }
     resource_->post_processor =
         std::make_shared<wenet::PostProcessor>(*post_process_opts_);
+
+    // Set decode opts
+    decode_options_->chunk_size = chunk_size_;
+    decode_options->num_left_chunks = num_left_chunks_;
```
pengzhendong, line 98: decode_options_

pengzhendong (changes requested):

runtime/core/api/wenet_api.cc
Mddct: done

runtime/binding/python/cpp/binding.cc
```diff
@@ -35,4 +35,6 @@ PYBIND11_MODULE(_wenet, m) {
   m.def("wenet_set_context_score", &wenet_set_context_score,
         "set context bonus score");
   m.def("wenet_set_language", &wenet_set_language, "set language");
+  m.def("wenet_set_chunk_size", &wenet_set_chunk_size, "set chunk size");
+  m.def("wenet_set_weights, &wenet_set_weights, "set decode weights");
```
pengzhendong, line 39: 漏了个引号

pengzhendong (changes requested):

runtime/binding/python/cpp/binding.cc
Mddct: done

pengzhendong (approved):

PR #1228, opened 2022-06-16
[bin] add comment for export_gpu_onnx.py

Mddct:


robin1001 (approved):

Mddct:
As @robin1001 said, the current target users of binding are novice users who are not familiar with speech recognition.
Binding currently hides cache logic, which can be confusing for beginners。

But I think we can provide a simple script example to get rid of dataloader like this:
```python
# Pseudo code for streamming
while：
  chunk = get_next_chunk_from_wav(chunk_size)

  output, att_cache, cnn_cache = model.forward_chunk(chunk, chunk_size, att_cache, cnn_cache)
  res = ctc_decoder(output)
```

Issue #1227, opened 2022-06-14
如何获得model


billqu01:
我通过aliyun用pip按照了wenet
 python -m pip install wenet -i https://mirrors.aliyun.com/pypi/simple/
我的程序如下，想试一试识别，但是没有下载到model，请问如何下载到model?
import sys
import wenet

def get_text_from_wav(dir, wav):
    model_dir = dir
    wav_file = wav
    decoder = wenet.Decoder(model_dir)
    ans = decoder.decode_wav(wav_file)
    print(ans)

# Press the green button in the gutter to run the script.
if __name__ == '__main__':
    print('test')

# See PyCharm help at https://www.jetbrains.com/help/pycharm/

robin1001:
@Mddct seems it is a strong requirement, when can we finish it?

Mddct:
> @Mddct seems it is a strong requirement, when can we finish it?

I‘m doing now ， maybe pr tomorrow

Mddct:
> 我通过aliyun用pip按照了wenet
>  python -m pip install wenet -i https://mirrors.aliyun.com/pypi/simple/
> 我的程序如下，想试一试识别，但是没有下载到model，请问如何下载到model?
> import sys
> import wenet
> 
> def get_text_from_wav(dir, wav):
>     model_dir = dir
>     wav_file = wav
>     decoder = wenet.Decoder(model_dir)
>     ans = decoder.decode_wav(wav_file)
>     print(ans)
> 
> # Press the green button in the gutter to run the script.
> if __name__ == '__main__':
>     print('test')
> 
> # See PyCharm help at https://www.jetbrains.com/help/pycharm/
> 

you can manually download mode from https://github.com/wenet-e2e/wenet/blob/main/docs/pretrained_models.md

future code will automatic download by language type

PR #1234, opened 2022-06-19
[binding] download model auto

Mddct:
##  https://github.com/wenet-e2e/wenet/issues/1227

### DIFF
--- a/runtime/binding/python/py/decoder.py
+++ b/runtime/binding/python/py/decoder.py
@@ -12,24 +12,11 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-import os
-import tarfile
-from pathlib import Path
 from typing import List, Optional
-from urllib.request import urlretrieve
 
 import _wenet
-import tqdm
 
-# TODO(Mddct): make assets class to support other language
-Assets = {
-    # wenetspeech
-    "chs":
-    "https://wenet-1256283475.cos.ap-shanghai.myqcloud.com/models/wenetspeech/20220506_u2pp_conformer_libtorch.tar.gz",
-    # gigaspeech
-    "en":
-    "https://wenet-1256283475.cos.ap-shanghai.myqcloud.com/models/gigaspeech/20210728_u2pp_conformer_libtorch.tar.gz"
-}
+from .hub import Hub
 
 
 class Decoder:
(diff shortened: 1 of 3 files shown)


runtime/binding/python/py/decoder.py
```diff
@@ -34,7 +50,11 @@ def __init__(self,
             context: context words
             context_score: bonus score when the context is matched
         """
+        if not model_dir:
```
robin1001, line 53: `model_dir is not None`, make it explicit.

runtime/binding/python/py/decoder.py
```diff
@@ -49,6 +69,61 @@ def reset(self):
         """ Reset status for next decoding """
         _wenet.wenet_reset(self.d)
 
+    def get_model_by_lang(self, lang: str):
+
+        assert lang in ['chs', 'en']
+        # NOTE(Mddct): model_dir structure
+        # Path.Home()/.went
+        # - chs
+        #    - 20220506_u2pp_conformer_libtorch
```
robin1001, line 78: we can just ignore the dirname `20220506_u2pp_conformer_libtorch` since we only have one model  for each language.

And please notes the files in the download dir are:
final.zip
units.txt
We have renamed words.txt to units.txt recently.

robin1001: It's better if we decouple the download logic to another module.

runtime/binding/python/setup.py
```diff
@@ -23,6 +23,7 @@ def cmake_extension(name, *args, **kwargs) -> setuptools.Extension:
 
 
 class BuildExtension(build_ext):
+
```
robin1001, line 26: please fallback the change here.

robin1001 (changes requested):
please see inline.

robin1001 (dismissed):

runtime/binding/python/py/hub.py
```diff
@@ -0,0 +1,86 @@
+import os
```
pengzhendong, line 1: Please complete the license.

pengzhendong (changes requested):

runtime/binding/python/py/hub.py
Mddct: done

pengzhendong (approved):

PR #1237, opened 2022-06-21
Model cache reuse in continuous decoding

zhuzizyf:
We should not reset the model, in order to reuse the cache.


pengzhendong:
We should reset the `encoder_outs_`.

robin1001:
we must reset the model to clear cache, reset encoder outputs, and offset.

PR #1277, opened 2022-07-03
[transducer] fix greedy search bug

Mddct:


wenet/transducer/transducer.py
```diff
@@ -216,27 +216,30 @@ def greedy_search(
...
         pred_out_step = None
-        per_frame_max_noblk = 5
+        per_frame_max_noblk = 100
         per_frame_noblk = 0
         while t < encoder_out_lens:
             encoder_out_step = encoder_out[:, t:t + 1, :]  # [1, 1, E]
-            if not prev_out_nblk:
-                pred_out_step, state_m, state_c = self.predictor.forward_step(
+            if prev_out_nblk:
+                pred_out_step, state_out_m, state_out_c = self.predictor.forward_step(
```
robin1001, line 226: We can encode the state into one tensor in the future so it be applied to any kind of predictors. Here `m` and `c` is for LSTM like predictor.

Mddct: yes， i will implement embedding predictor after aishell works， and then concat into one state

robin1001 (approved):

PR #1283, opened 2022-07-05
［transducer］ search: beam and greedy

Mddct:


Mddct:
close now， will try more decode method

PR #1289, opened 2022-07-07
[transducer] track aishell results

Mddct:
track the results of transducer not for merge now


robin1001 (dismissed):

Mddct:
It seems that multiple loss functions work ok, next we will implement the embedding predictor to see the effect to improve the usage of gpu and reduce the size of the model

robin1001 (approved):

robin1001:
Let's merge it and do the above optimization in future PRs.

PR #1294, opened 2022-07-11
[transducer] transducer embedding predictor

Mddct:
TODO:
- [x] embedding predictor
- [x] tie embedding
- [x] aishell results using embedding predictor


wenet/transducer/example_embedding_predictor.yaml
```diff
@@ -0,0 +1,96 @@
+encoder: conformer
```
robin1001, line 1: please move the file to aishell/rnnt/conf

Mddct:
issues： we should not tie blank when tie embedding ，easy to cause model collapse

Mddct:
> issues： we should not tie blank when tie embedding ，easy to cause model collapse

this is not a problem, due to code bug,  And i'll paste aishell results later which can achieve the same performace as rnn or even better

Mddct:
@robin1001  I remove tie embedding. Maybe I'll re-add it if i find it can improve the accuracy. 
for now:

aishell rnnt-greedy search: 4.892 
160 epoch avg10

I think  it's time to merge it now,

TODO: (next stage)
- [ ]  embedding + conv
- [ ]  this version of embedding + conv

robin1001:
Great job! Let's keep it simple and stupid now. Looking forward to your future work.

Mddct:
WIP， Looks like the conflict's gone, but I need to double check the decoding

Mddct:
> WIP， Looks like the conflict's gone, but I need to double check the decoding

works  ok

robin1001 (approved):

PR #1302, opened 2022-07-14
Zhyyao beamsearch

zhyyao:
I have found that RNN-Transducer prefix beam search doesn't need to record last output, so I fix this which make the beam search  wer change from 4.80 to 4.67


xingchensong:
Consider removing `last` in `class Sequence()` ? seems it is not required now.

Mddct:
if prefix beam search from 4.80 to 4.67, Will rescoer get any better？

robin1001 (approved):

PR #1309, opened 2022-07-16
Add ctc beam search with transducer and attention rescore

zhyyao:
I got 4.54 for this decoding method with the model greedy search score 4.88. And also split a large function code to some part.


wenet/transducer/search/prefix_beam_search.py
```diff
@@ -183,3 +208,88 @@ def prefix_beam_search(
             fusion_A.sort(key=lambda x: x.score, reverse=True)
             beam_init = fusion_A[:beam_size]
         return beam_init, encoder_out
+
+    def ctc_prefix_beam_search(
```
robin1001, line 212: what's the difference of the CTC prefix beam search to corresponding in ASRModel?

zhyyao: I have changed the transducer inherit class to `wenet.transformer.asr_model` and delete the ctc_prefix_beam_search from `wenet.tranducer.search.prefix_beam_search`

robin1001 (approved):

PR #1363, opened 2022-08-07
[runtime] fix runtime structure

pengzhendong:


robin1001:
Why change all directories to low case? It's better we keep it unchanged. GPU/IPU/XPU/BPU are in upper case, typically.

pengzhendong:
> Why change all directories to low case? It's better we keep it unchanged. GPU/IPU/XPU/BPU are in upper case, typically.

Because `android` is low case, and I think the low case is convenient to use `cd` command. We can create the `README.md` under each folder to explain the meaning of GPU/IPU/XPU/BPU.

robin1001:
okay!

robin1001 (approved):

robin1001:
okay!

PR #1399, opened 2022-08-23
Delete file before update of symlink

nshmyrev:
Currently if file exists it issues warning in the log


nshmyrev:
And it doesn't update symlink so continues with the old model

robin1001 (approved):

PR #1406, opened 2022-08-26
Fix GPU Runtime build error

azraelkuan:
when building gpu triton server docker, i met a problem:
```
W: GPG error: https://developer.download.nvidia.com/compute/cuda/repos/ubuntu2004/x86_64  InRelease: The following signatures couldn't be verified because the public key is not available: NO_PUBKEY A4B469963BF863CC
E: The repository 'https://developer.download.nvidia.com/compute/cuda/repos/ubuntu2004/x86_64  InRelease' is not signed.
```
so i add a new line in the dockerfile:
`RUN apt-key adv --fetch-keys https://developer.download.nvidia.com/compute/cuda/repos/ubuntu1804/x86_64/3bf863cc.pub`

### DIFF
--- a/runtime/gpu/Dockerfile/Dockerfile.client
+++ b/runtime/gpu/Dockerfile/Dockerfile.client
@@ -2,6 +2,7 @@ FROM  nvcr.io/nvidia/tritonserver:22.03-py3-sdk
 LABEL maintainer="NVIDIA"
 LABEL repository="tritonserver"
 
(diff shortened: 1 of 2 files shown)


robin1001:
@yuekaizhang please follow the PR.

yuekaizhang:
Hi, thanks for the fix. We would like use tritonserver 22.07 image later which has no GPG error.

yuekaizhang (approved):

PR #1418, opened 2022-08-30
fix bug: aishell2 word_segmentation.py

LvHang:
The output of ```awk '{$1,99}'``` should be like ```xxx 99```.


robin1001 (approved):

robin1001:
First blood on WeNet?

PR #1419, opened 2022-08-31
[runtime] support read fst using mmap

mackong:
# Background
## the fst file
```
➜  lm git:(standalone) ll lm.fst
-rw-r--r-- 1 mackong mackong 12G Aug  3 10:09 lm.fst
```
## disk
```
➜  ~ lspci | grep SSD
04:00.0 Non-Volatile memory controller: SK hynix Gold P31 SSD
```

# READ mode
when use *READ* mode, we got the following performance
```
I0831 22:01:58.278613 14004 decoder_main.cc:180] Total: decoded 9550158ms audio taken 3076034ms.
I0831 22:01:58.278636 14004 decoder_main.cc:182] RTF: 0.3221
```
and the resource usage is
read

# MMAP mode
when use *MMAP* mode, we got the following performance
```
I0831 20:42:52.785739  8840 decoder_main.cc:180] Total: decoded 9550158ms audio taken 2995321ms.
I0831 20:42:52.785763  8840 decoder_main.cc:182] RTF: 0.3136
```
and the resource usage is
mmap

### DIFF
--- a/runtime/core/decoder/params.h
+++ b/runtime/core/decoder/params.h
@@ -47,6 +47,7 @@ DEFINE_int32(sample_rate, 16000, "sample rate for audio");
 
 // TLG fst
 DEFINE_string(fst_path, "", "TLG fst path");
+DEFINE_bool(fst_use_mmap, false, "read TLG fst use mmap or not");
 
 // DecodeOptions flags
 DEFINE_int32(chunk_size, 16, "decoding chunk size");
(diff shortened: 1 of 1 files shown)


robin1001:
Any performance benchmark?

mackong:
> Any performance benchmark?

Updated a rough benchmark.

robin1001:
There is no gain in our evaluation with smaller FST, I think it's better if we use ConstFst directly, please see https://groups.google.com/g/kaldi-help/c/4Sqw6vKu9CI

mackong:
Yes, for small FST, no need to use mmap for memory save.

PR #1432, opened 2022-09-05
[dataset] add simple rir perturb

Mddct:
copy some code from scipy to support rir convolve

Note: torchaudio also supplies convolve in  prototype: https://pytorch.org/audio/master/prototype.functional.html
when stable, we can replace fft_convolve with  torchaudio.convolve

### DIFF
--- a/wenet/bin/recognize.py
+++ b/wenet/bin/recognize.py
@@ -160,6 +160,7 @@ def main():
     test_conf['filter_conf']['max_output_input_ratio'] = 102400
     test_conf['filter_conf']['min_output_input_ratio'] = 0
     test_conf['speed_perturb'] = False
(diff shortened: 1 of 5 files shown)


robin1001:
How about the implementation in wespeaker https://github.com/wenet-e2e/wespeaker/blob/master/wespeaker/dataset/processor.py#L301?

Mddct:
> How about the implementation in wespeaker https://github.com/wenet-e2e/wespeaker/blob/master/wespeaker/dataset/processor.py#L301?
> 1. scipy.signal is used for convolution.
> 2. LMDB is used as the storage system for NOISE and RIR.


Lmdb may be not a good choice for large
  scale noisy resources eg: one millions noise files and much more room

PR #1476, opened 2022-10-03
refactor(onnx): [cr_id_skip] Export dynamic quantized onnx model

xingchensong:
ebfbe5f18e88aec83c2fd214985eefc

### DIFF
--- a/wenet/bin/export_onnx_cpu.py
+++ b/wenet/bin/export_onnx_cpu.py
@@ -29,6 +29,7 @@ from wenet.utils.init_model import init_model
 try:
     import onnx
     import onnxruntime
+    from onnxruntime.quantization import quantize_dynamic, QuantType
 except ImportError:
     print('Please install onnx and onnxruntime!')
     sys.exit(1)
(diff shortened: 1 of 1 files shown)


wenet/bin/export_onnx_cpu.py
```diff
@@ -169,6 +170,11 @@ def export_encoder(asr_model, args):
     #   the file and resave it.
     onnx.save(onnx_encoder, encoder_outpath)
     print_input_output_info(onnx_encoder, "onnx_encoder")
+    # Dynamic quantization
+    model_fp32 = encoder_outpath
+    model_quant = os.path.join(args['output_dir'], 'encoder.quant.onnx')
+    quantized_model = quantize_dynamic(
```
pengzhendong, line 176: I think we do need the returned value.

pengzhendong (changes requested):

wenet/bin/export_onnx_cpu.py
xingchensong: done.

pengzhendong (approved):

PR #1479, opened 2022-10-04
fix(wenetspeech): Fix dir name

xingchensong:


xingchensong:
$feat_dir never defined.

robin1001 (approved):

PR #1504, opened 2022-10-19
blockformer

LeonWlw:
This PR is about implementation of blockformer in WeNet.
(Original paper: [https://arxiv.org/abs/2207.11697](https://arxiv.org/abs/2207.11697))
- Implementation Details
    - add se layer ensemble conformer encoder outputs
    - add se layer ensemble transformer decoder outputs
    - using relative positional encoding in decoder
In main branch, extracting features by torchaudio make a little worse results than the paper. I will push a branch using kaldi features for aishell recipe which can reproduce results in the paper.

### DIFF
--- a/wenet/transformer/se_layer.py
+++ b/wenet/transformer/se_layer.py
@@ -1,0 +1,37 @@
+# Copyright (c) 2022 Mininglamp Com (Liuwei Wei, Xiaoming Ren)
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# Modified from ESPnet(https://github.com/espnet/espnet)
+
+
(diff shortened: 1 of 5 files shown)


robin1001:
I think it's better if we add the experiment results on AIShell-1 and LibriSpeech, to show that we can get consistent and solid gain by using the model.

wenet/transformer/se_layer.py
```diff
@@ -0,0 +1,37 @@
...
+        self.fc = torch.nn.Sequential(
+            torch.nn.Linear(channel, channel // reduction, bias=False),
+            torch.nn.ReLU(inplace=True),
+            torch.nn.Linear(channel // reduction, channel, bias=False),
+            torch.nn.Sigmoid()
+        )
+
+    def forward(self, x: torch.Tensor) -> torch.Tensor:
+        b, c, _, _ = x.size()
+        y = self.avg_pool(x).view(b, c)
```
kobenaxie, line 34: avg_pool over T and D dim should consider pad_mask ?

LeonWlw:
> I think it's better if we add the experiment results on AIShell-1 and LibriSpeech, to show that we can get consistent and solid gain by using the model.

@robin1001 results of aishell has been added

wenet/transformer/se_layer.py
lezasantaizi: thanks for your remind, we will update pad_mask to the code and retrain it .

903859154:
hi, i run blockformer in 3080 and it just used 30% - 40% gpu. I change batchsize bigger and numworker but it didn't work. So what should i do to take more use of gpu?

PR #1521, opened 2022-10-26
add hlg decode

aluminumbox:
We add hlg decode using k2. Now we can use k2 to compile a HLG, and decode using python with cuda.
We provide two decoding algorithm. One is onebest, which is same as ctc_prefix_beam_search with lm score, another is attention rescore.
Notice that in attention rescore, we have 3 different scale parameter, which are lm_scale, decoder_scale and r_decoder_scale.
Special thanks to K2 groups great work!

### DIFF
--- a/examples/aishell/s0/README.md
+++ b/examples/aishell/s0/README.md
@@ -19,7 +19,7 @@
 
 * Feature info: using fbank feature, dither=1.0, cmvn, oneline speed perturb
 * Training info: lr 0.001, batch size 16, 8 gpu, acc_grad 1, 360 epochs
-* Decoding info: ctc_weight 0.3, reverse_weight 0.5  average_num 30
+* Decoding info: ctc_weight 0.3, reverse_weight 0.5  average_num 30, lm_scale 0.7, decoder_scale 0.1, r_decoder_scale 0.7
 * Git hash: 5a1342312668e7a5abb83aed1e53256819cebf95
 
 | decoding mode/chunk size  | full  | 16    |
@@ -28,6 +28,8 @@
 | ctc prefix beam search    | 5.17  | 5.81  |
 | attention rescoring       | 4.63  | 5.05  |
 | LM + attention rescoring  | 4.40  | 4.75  |
+| HLG(k2 LM)                | 4.81  | 5.27  |
+| HLG(k2 LM)  + attention rescoring | 4.32  | 4.70  |
 
 ## Unified Conformer Result
 
--- a/examples/aishell/s0/run.sh
+++ b/examples/aishell/s0/run.sh
@@ -239,4 +239,50 @@ if [ ${stage} -le 7 ] && [ ${stop_stage} -ge 7 ]; then
   # Please see $dir/lm_with_runtime for wer
 fi
 
+# Optionally, you can decode with k2 hlg
+if [ ${stage} -le 8 ] && [ ${stop_stage} -ge 8 ]; then
+  if [ ! -f data/local/lm/lm.arpa ]; then
+    echo "Please run prepare dict and train lm in Stage 7" || exit 1;
+  fi
+
+  # 8.1 Build decoding HLG
+  required="data/local/hlg/HLG.pt data/local/hlg/words.txt"
+  for f in $required; do
+    if [ ! -f $f ]; then
+      tools/k2/make_hlg.sh data/local/dict/ data/local/lm/ data/local/hlg
+      break
+    fi
+  done
+
+  # 8.2 Decode using HLG
+  decoding_chunk_size=
+  lm_scale=0.7
+  decoder_scale=0.1
+  r_decoder_scale=0.7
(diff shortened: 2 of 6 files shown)


wenet/transformer/asr_model.py
```diff
@@ -20,6 +20,10 @@
 
 from torch.nn.utils.rnn import pad_sequence
 
+import k2
```
robin1001, line 23: move it into `hlg_onebest` & `hlg_rescore`, so we can run it without k2 & icefall.

examples/aishell/s0/README.md
```diff
@@ -28,6 +28,8 @@
 | ctc prefix beam search    | 5.17  | 5.81  |
 | attention rescoring       | 4.63  | 5.05  |
 | LM + attention rescoring  | 4.40  | 4.75  |
+| HLG                       | 4.81  | 5.27  |
```
robin1001, line 31: How about HLG(k2 LM), so it is easy to understand for the users.

robin1001 (changes requested):
please see inline.

csukuangfj:
What is the motivation to add HLG from k2?

wenet/transformer/asr_model.py
aluminumbox: To avoid duplicate code, we use move these import into try except

examples/aishell/s0/README.md
aluminumbox: Done

aluminumbox:
> What is the motivation to add HLG from k2?

k2 can do batch decoding with cuda, and it has python interface. 
It is easier to use and faster compared with wenet runtime tlg

csukuangfj:
> > What is the motivation to add HLG from k2?
> 
> k2 can do batch decoding with cuda, and it has python interface. It is easier to use and faster compared with wenet runtime tlg

Thanks! Do you have some benchmarks to share?

aluminumbox:
> 



> > > What is the motivation to add HLG from k2?
> > 
> > 
> > k2 can do batch decoding with cuda, and it has python interface. It is easier to use and faster compared with wenet runtime tlg
> 
> Thanks! Do you have some benchmarks to share?

The wer performance is already compared in README. For speed comparision, wenet runtime default using batchsize=1, which is not comparable with hlg decode using batchsize=16. I will do the speed benchmark later in a fair scenario

robin1001:
It is more convienent to use FSA in python than c++ in openfst, especially for the beginners. And Dan thinks it is okay to support k2 in wenet.

robin1001:
Besides, we can use k2 with GPU.

examples/aishell/s0/run.sh
```diff
@@ -239,4 +239,44 @@ if [ ${stage} -le 7 ] && [ ${stop_stage} -ge 7 ]; then
...
 fi
 
+# Optionally, you can decode with k2 hlg
+if [ ${stage} -le 8 ] && [ ${stop_stage} -ge 8 ]; then
+  if [ ! -f data/local/lm/lm.arpa ]; then
+    echo "Please run prepare dict and train lm in Stage 7"
+  fi
+
+  # 8.1 Build decoding HLG
+  tools/k2/make_hlg.sh data/local/dict/ data/local/lm/ data/local/hlg
```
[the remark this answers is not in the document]
aluminumbox: Done

examples/aishell/s0/run.sh
```diff
@@ -239,4 +239,44 @@ if [ ${stage} -le 7 ] && [ ${stop_stage} -ge 7 ]; then
   # Please see $dir/lm_with_runtime for wer
 fi
 
+# Optionally, you can decode with k2 hlg
+if [ ${stage} -le 8 ] && [ ${stop_stage} -ge 8 ]; then
+  if [ ! -f data/local/lm/lm.arpa ]; then
+    echo "Please run prepare dict and train lm in Stage 7"
```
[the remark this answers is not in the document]
aluminumbox: Done

robin1001 (approved):

zw76859420:
> We add hlg decode using k2. Now we can use k2 to compile a HLG, and decode using python with cuda. We provide two decoding algorithm. One is onebest, which is same as ctc_prefix_beam_search with lm score, another is attention rescore. Notice that in attention rescore, we have 3 different scale parameter, which are lm_scale, decoder_scale and r_decoder_scale. Special thanks to K2 groups great work!

Thank you for sharing, but we would like to know how we can access HLG Decoding using c++？

zw76859420:
It‘s unpossible that we load the dynamic libraries provided by k2 and wenet at the same time.
So, can we combine both wenet and k2 in a simple way?

csukuangfj:
> It‘s unpossible that we load the dynamic libraries provided by k2 and wenet at the same time.

Why is it impossible?

Have you used the same version of PyTorch and CUDA to compile k2 and went?

PR #1533, opened 2022-11-02
iOS porting: Add iOS cmake files, LibTorch library and test application.

Ma-Dan:


robin1001:
Great job. 

It seems all the source code of Libtorch are added, which is not prefered. We should add them as the third_party library.

robin1001:
for cleaning up LibTorch files, please see https://pytorch.org/tutorials/recipes/model_preparation_ios.html#add-the-model-and-pytorch-library-on-ios.

Issue #1529, opened 2022-10-31
Cannot use Context in Librispeech model


mapledxf:
Hi there,
Thanks for the awesome project. Now I am facing some Context related issue here:

First, I registered a context word like "TURN" which is not in the units.txt coming from the librispeech model, but there is a space-prefixed word "▁TURN 4622" in the units.txt.

So when I run the code, the context word "TURN" has been split into 2 ids: 4619 2930 which are "TRU" and "N" respectively. And in the decoding phase, the decoder will return the "▁TURN" id which is 4622, and that will cause the Context cannot be matched.

Any idea how I can solve that?

robin1001:
Yes, there is some bug for English if you use BPE as modeling unit. @pengzhendong do you have time to fix it?

pengzhendong:
> Yes, there is some bug for English if you use BPE as modeling unit. @pengzhendong do you have time to fix it?

I will fix it.

mapledxf:
@pengzhendong 
So instead of calling SplitUTF8StringToWords, maybe we should use BPE to decode the text into words for English inputs.

robin1001:
I think we can just add '▁' at the begin of English word, then use the same code as before.

robin1001:
Fixed, please see https://github.com/wenet-e2e/wenet/pull/1540.

PR #1540, opened 2022-11-05
[runtime] fix english context #1529

pengzhendong:


runtime/core/utils/string.cc
```diff
@@ -131,6 +140,16 @@ bool SplitUTF8StringToWords(
...
         word += chars[i];
       }
+      // Skip space.
+      if (word == " ") {
+        start = end;
+        continue;
+      }
+      // Add '▁' at the beginning of English word.
+      if (IsAlpha(word)) {
+        word = "▁" + word;
```
robin1001, line 150: kSpaceSymbol

robin1001: What if we use char as English modeling unit?

robin1001 (changes requested):
please see inline.

robin1001 (approved):

PR #1544, opened 2022-11-07
[wenet/squeezeformer] compat with websocket server

yygle:


veelion:
Great job! test for chunk size -1 works fine, works both for cpu and gpu.

robin1001 (approved):

PR #1610, opened 2022-12-11
[RNN-T] fix the rnnt joint post fnn shape bug

Honei:
The RNN-T joint post_ffn shape `self.post_ffn = nn.Linear(enc_output_size, join_dim)` is not match the `out` variable.

I check the the correct shape is `self.post_ffn = nn.Linear(join_dim, join_dim)`


Mddct:
postjoin_linear means both encoder and predictor dim should be equal to encdoer_output_size,
encoder -> [B,T,d]
predictor-> [B,T,d]

Honei:
so if set the `postjoin_linear = True`, the `prejoin_linear` must be `False`。
The postjoin_linear process the `out` and the shape of `out` is `[B T U V]`

PR #1670, opened 2023-01-29
add http server/client

aluminumbox:
add a http server/client for those who prefer using http request


robin1001 (changes requested):
Start/End is required in WebSocket since it's a dual mode streaming application. HTTP is single request/response, I think we can put all parameters including the pcm data in HTTP body. So we can keep it as simple as possible.

robin1001:
Maybe we can refer https://github.com/Baidu-AIP/speech-demo/blob/master/rest-api-asr/python/asr_json.py#L98.

aluminumbox:
> Start/End is required in WebSocket since it's a dual mode streaming application. HTTP is single request/response, I think we can put all parameters including the pcm data in HTTP body. So we can keep it as simple as possible.



> Start/End is required in WebSocket since it's a dual mode streaming application. HTTP is single request/response, I think we can put all parameters including the pcm data in HTTP body. So we can keep it as simple as possible.

parameters are in string format, should be put in header. can not mix with binary data buffer body

aluminumbox:
> Maybe we can refer https://github.com/Baidu-AIP/speech-demo/blob/master/rest-api-asr/python/asr_json.py#L98.

use base64 encode binary data increase body size, and waste bandwidth. beast already provide buffer body.

aluminumbox:
> Start/End is required in WebSocket since it's a dual mode streaming application. HTTP is single request/response, I think we can put all parameters including the pcm data in HTTP body. So we can keep it as simple as possible.

moreover, buffer body is not single request/response, this is streaming mode, you can send data in streaming way

aluminumbox:
changed to non-streaming http, keep the streaming http commit in case someone needs it

runtime/core/http/http_client.cc
```diff
@@ -0,0 +1,65 @@
...
+void HttpClient::Connect() {
+  tcp::resolver resolver{ioc_};
+  // Look up the domain name
+  auto const results = resolver.resolve(hostname_, std::to_string(port_));
+  stream_.connect(results);
+}
+
+void HttpClient::SendBinaryData(const void* data, size_t size) {
+  try {
+    json::value start_tag = {{"signal", "start"},
```
robin1001, line 43: signal is not required for http

runtime/core/http/http_server.cc
```diff
@@ -0,0 +1,206 @@
...
+  // Send a TCP shutdown
+  socket_.shutdown(tcp::socket::shutdown_send, ec_);
+}
+
+void ConnectionHandler::OnText(const std::string& message) {
+  LOG(INFO) << message;
+  json::value v = json::parse(message);
+  if (v.is_object()) {
+    json::object obj = v.get_object();
+    if (obj.find("signal") != obj.end()) {
```
robin1001, line 137: singal start and end are not required here.

aluminumbox: deleted signal start and end message

robin1001 (approved):

Issue #1690, opened 2023-02-10
Streaming Efficient Conformer isn't working


Uncomfy:
I was able to reproduce the Efficient Conformer model using the scripts provided in the AIShell example. If I run the decoding using "recognize.py" with "decoding_chunk_size" set to 16 I'm able to reproduce the numbers reported by you, but if I add "simulate_streaming" flag or try to use this model in the runtime - CER jumps to almost 50%.

Is Efficient Conformer just not designed for streaming?

robin1001:
@zwglory , please follow the issue.

zwglory:
If you want a streaming model, `causal: true` needs to be specified when training the model.

In addition, `decoding_chunk_size ` is best set to a common multiple of `group_size`. If group_size=3, then `decoding_chunk_size` can be 18.


```
encoder: efficientConformer
encoder_conf:
    activation_type: 'swish'
    attention_heads: 8
    causal: true
    cnn_module_kernel: 15
    cnn_module_norm: 'layer_norm'
    dropout_rate: 0.1
    input_layer: conv2d
    linear_units: 2048
    normalize_before: true
    num_blocks: 12
    output_size: 256
    pos_enc_layer_type: 'rel_pos'
    attention_dropout_rate: 0.1
    positional_dropout_rate: 0.1
    use_cnn_module: true
    use_dynamic_chunk: true
    use_dynamic_left_chunk: false
    efficient_conf:
        stride_layer_idx: [3]           # layer id with StrideConv
        stride: [2]                     # stride size of each StrideConv
        group_layer_idx: [0, 1, 2, 3]   # layer id with GroupedAttention
        group_size: 3                   # group size of every GroupedAttention layer
        stride_kernel: true             # true: recompute cnn kernels with stride
```

Uncomfy:
Thank you, will try to train a causal model now :)

Uncomfy:
Results for causal Efficient Conformer v2 (attention rescoring):
Full audio: 4.83%
Decoding chunk 16, "simulate_streaming" is enabled: 5.81%

Cannot decode with chunk of size 18 (or any chunk divisible by 3, actually), because I get the following error:
<details>
  <summary>error</summary>

  ```
Traceback (most recent call last):
  File ".../wenet_ec/examples/aishell/s0/wenet/bin/recognize.py", line 357, in <module>
    main()
  File ".../wenet_ec/examples/aishell/s0/wenet/bin/recognize.py", line 313, in main
[...]

PR #1701, opened 2023-02-21
[Efficient Conformer] Support ONNX GPU export, add librispeech results, and fix V2 streaming decode issue

zwglory:
Support ONNX GPU export.

- After exporting onnx, the loss of CER is small. Using the `train_u2++_efficonformer_v1_stream.yaml` with `causal:true`, the CER increases from 9.30% to 9.33% in our dataset.

Add librispeech results and conf:

- train_u2++_efficonformer_v1.yaml
- train_u2++_efficonformer_v2.yaml

Add streaming conf in aishell: 

- train_u2++_efficonformer_v1_stream.yaml

Fix bug of V2 streaming decode: 

- Fix issue https://github.com/wenet-e2e/wenet/issues/1690#issuecomment-1434867574


wenet/efficient_conformer/encoder.py
```diff
@@ -381,32 +396,42 @@ def forward_chunk(
...
+                att_cache_trunc = xs.size(1) + \
+                    att_cache.size(2) // factor - pos_emb.size(1) + 1
             xs, _, new_att_cache, new_cnn_cache = layer(
                 xs, att_mask, pos_emb,
                 mask_pad=mask_pad,
-                att_cache=att_cache[i:i + 1, :, ::factor, :],
+                att_cache=att_cache[i:i + 1, :, ::factor, :][:, :, att_cache_trunc:, :],
                 cnn_cache=cnn_cache[i, :, :, :]
                 if cnn_cache.size(0) > 0 else cnn_cache
             )
```
xingchensong, line 437: Q1: 

```
xs.size(1) + att_cache.size(2) / factor > pos_emb.size(1)
```
OR
```
( xs.size(1) + att_cache.size(2) ) / factor > pos_emb.size(1)
```
?
results of the above two lines are not equal :

image


Q2:

What do mean by `double the chunk_size` ? I think `[:, :, att_cache_trunc:, :]` will simply eliminate any unnecessary attention cache information at the beginning, so where is the `double` ?

examples/librispeech/s0/README.md
```diff
@@ -219,3 +219,53 @@ test other
...
+| attention rescoring    | 3.17 | 3.44 | 3.41 |
+
+test other
+
+| decoding mode          | full | 18    | 16    |
+|------------------------|------|-------|-------|
+| attention decoder      | 8.51 | 9.24  | 9.25  |
+| ctc_greedy_search      | 8.94 | 10.04 | 10.06 |
+| ctc prefix beam search | 8.91 | 10    | 10.01 |
+| attention rescoring    | 8.21 | 9.25  | 9.25  |
```
xingchensong, line 246: Thx, the results are much better than the standard conformer! I would suggest adding model params in README for a clearer comparison.

wenet/efficient_conformer/encoder.py
```diff
@@ -355,22 +364,28 @@ def forward_chunk(
...
+        chunk_masks = chunk_masks.unsqueeze(1)  # (1, 1, xs-time)
+
+        real_len = 0
+        if self.global_chunk_size > 0:
+            # for ONNX decode simulation， padding xs to chunk_size
+            real_len = xs.size(1)
+            pad_len = self.chunk_feature_map - real_len
+            xs = F.pad(xs, (0, 0, 0, pad_len), value=0.0)
+            chunk_masks = F.pad(chunk_masks, (0, pad_len), value=0.0)
+
```
xingchensong, line 380: Out of curiosity, will padding only be applied to the last chunk, given that previous chunks always have a valid chunk size?

wenet/efficient_conformer/encoder.py
zwglory: Q1: `xs.size(1) + att_cache.size(2) / factor > pos_emb.size(1)`, because `xs` was downsampled in the previous block.

Q2: For the `train_u2++_efficonformer_v2.yaml`, downsample rate is 1/2 (conv2d2) + 1/4 (efficonformer block), so 18->36 can reduce the downsampling loss.

The description of double is really inaccurate here, so let me adjust it.

examples/librispeech/s0/README.md
zwglory: OK, no problem

wenet/efficient_conformer/encoder.py
zwglory: Yes, padding will only be applied to the last chunk. Also, this part is currently only valid if you manually specify `use_onnx=True`, to simulate the CER after exporting ONNX.

xingchensong (approved):

xingchensong:
Many thx!

PR #1725, opened 2023-03-09
add LFMMI loss

aluminumbox:
add LFMMI loss, add torch.jit.ignore for lfmmi function

### DIFF
--- a/wenet/transformer/asr_model.py
+++ b/wenet/transformer/asr_model.py
@@ -24,9 +24,12 @@ try:
     import k2
     from icefall.utils import get_texts
     from icefall.decode import get_lattice, Nbest, one_best_decoding
+    from icefall.mmi import LFMMILoss
+    from icefall.mmi_graph_compiler import MmiTrainingGraphCompiler
 except ImportError:
     print('Failed to import k2 and icefall. \
-        Notice that they are necessary for hlg_onebest and hlg_rescore')
+        Notice that they are necessary for \
+        hlg_onebest/hlg_rescore decoding and LF-MMI training')
 
 from wenet.transformer.ctc import CTC
 from wenet.transformer.decoder import TransformerDecoder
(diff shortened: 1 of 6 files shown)


wenet/transformer/asr_model.py
```diff
@@ -89,6 +94,9 @@ def forward(
             text: (Batch, Length)
             text_lengths: (Batch,)
         """
+        if self.lfmmi_dir != '':
```
robin1001, line 97: Why load it in forward? I think we should load it in construction.

robin1001: Then there is no need to use hasattr when loading resource

wenet/transformer/asr_model.py
```diff
@@ -552,6 +563,54 @@ def attention_rescoring(
...
+            with open('{}/words.txt'.format(self.lfmmi_dir), 'r') as fin:
+                for line in fin:
+                    arr = line.strip().split()
+                    assert len(arr) == 2
+                    self.word_table[int(arr[1])] = arr[0]
+
+    @torch.jit.ignore(drop=True)
+    def _calc_lfmmi_loss(self, encoder_out, encoder_mask, text):
+        ctc_probs = self.ctc.log_softmax(encoder_out)
+        supervision_segments = torch.stack(
```
robin1001, line 599: why we should move it to cpu?

wenet/transformer/asr_model.py
aluminumbox: > Why load it in forward? I think we should load it in construction.

because i need to decorate it with torch.jit.ignore, i dont want to decorate whole forward function with torch.jit.ignore

wenet/transformer/asr_model.py
aluminumbox: k2 supervision_segments  on cpu

wenet/transformer/asr_model.py
aluminumbox: now we move load_mmi_resource into __init__ and remove hasattr judgement

robin1001 (approved):

kli017:
@aluminumbox 请问现在lfmmi在runtime推理中能够使用吗

PR #1729, opened 2023-03-10
Update wav.h

ferb2015:
If the file is in text format, it will loop on `while (0 != strncmp(header.data, "data", 4))` , so we need to add a judgment whether it is an audio file.

### DIFF
--- a/runtime/core/frontend/wav.h
+++ b/runtime/core/frontend/wav.h
@@ -71,6 +71,13 @@ class WavReader {
 
     WavHeader header;
     fread(&header, 1, sizeof(header), fp);
+    if (!(header.riff[0] == 'R' && header.riff[1] == 'I' && header.riff[2] == 'F' && header.riff[3] == 'F' && \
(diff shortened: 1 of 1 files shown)


robin1001:
please fix the lint.

runtime/core/frontend/wav.h
```diff
@@ -71,6 +71,13 @@ class WavReader {
 
     WavHeader header;
     fread(&header, 1, sizeof(header), fp);
+    if (!(header.riff[0] == 'R' && header.riff[1] == 'I' && header.riff[2] == 'F' && header.riff[3] == 'F' && \
```
robin1001, line 74: You can use `strncmp` to compare it with 'RIFF'

robin1001:
runtime/core/frontend/wav.h:74:  Lines should be <= 80 characters long  [whitespace/line_length] [2]
runtime/core/frontend/wav.h:74:  Missing space before ( in if(  [whitespace/parens] [5]

robin1001 (approved):

PR #1734, opened 2023-03-10
CIF Related

MrSupW:
Implement CIF related functions and complete baseline testing on AISHELL.


examples/aishell/cif/local/aishell_data_prep.sh
```diff
@@ -0,0 +1,65 @@
+#!/bin/bash
```
robin1001, line 1: the files in local is the same with ../s0, right? If yes, please use soft link.

robin1001:
@MrSupW Thanks for the contribution, it's really a big and amazing work! Since it's a lot of change, it will take several days to finish the whole review pipeline.

And please fist make sure:
* all the lint is paased, so we get tidy and good style code.
* please check if we could share some code with the preivous implementation, so we can reuse the common code blocks as much as possible.

PR #1738, opened 2023-03-12
CIF Related

MrSupW:
Complete the implementation of CIF-related code, including two types of Predictor and two types of CIF-Decoder. Meanwhile,  four baseline experiments were conducted on the AISHELL dataset.


wenet/cif/embedding.py
```diff
@@ -0,0 +1,100 @@
+# Copyright (c) 2023 ASLP@NWPU (authors: He Wang, Fan Yu)
```
robin1001, line 1: The embedding.py shares almost the same code with transformer/embedding.py, I think we can reuse it.

wenet/cif/attention.py
```diff
@@ -0,0 +1,291 @@
+# Copyright (c) 2023 ASLP@NWPU (authors: He Wang, Fan Yu)
```
robin1001, line 1: same as embedding.py, attention shares almost the same code as transformer/attention.py, and transformer/attention.py takes streamming into account.

wenet/cif/utils.py
```diff
@@ -0,0 +1,102 @@
...
+# FunASR(https://github.com/alibaba-damo-academy/FunASR)
+
+from typing import Optional
+
+import six
+import torch
+import numpy as np
+
+
+def make_pad_mask(lengths: torch.Tensor, length_dim: int = -1,
```
robin1001, line 23: this function is defined in `utils/mask.py`

wenet/cif/utils.py
```diff
@@ -0,0 +1,102 @@
...
+
+    seq_range = torch.arange(0, maxlen, dtype=torch.int64)
+    seq_range_expand = seq_range.unsqueeze(0).expand(bs, maxlen)
+    seq_length_expand = lengths.unsqueeze(-1).to(seq_range_expand.device)
+    mask = seq_range_expand >= seq_length_expand
+
+    return mask
+
+
+def sequence_mask(lengths, maxlen: Optional[int] = None,
```
robin1001, line 58: this function is almost the same as subsequent_mask in utils/mask.py

robin1001 (approved):

PR #1761, opened 2023-03-19
[fix] fix index error

lucy9527:
There may cause eoor when length of y is 1.


robin1001 (approved):

robin1001:
please fix the lint problem.

PR #1762, opened 2023-03-20
[transformer] remove concat after to simplify the code flow

robin1001:


robin1001:
We can still:

- [x] load the previous pretrained model.
- [x] decode works

image

robin1001:
@zwglory @yygle please pay attention to the PR, I think we can remove concat after in squeezeformer and efficient conformer.

robin1001:
> https://github.com/wenet-e2e/wenet/blob/4467198fd03a77aca55bffd8b987686acc9f1205/wenet/bin/export_onnx_bpu.py#L545
> 
> 这行是不是要删掉

后面分模块去删除

PR #1766, opened 2023-03-20
[wenet] remove typeguard since we use jit for type check and inference

robin1001:


robin1001:
And there is no `check_argument_types` in python 3.9+.

PR #1784, opened 2023-03-27
[cif] use standard TransformerDecoder instead of CIFDecoder

robin1001:


MrSupW (approved):
No problem. It's all good

PR #1800, opened 2023-03-31
Added post-processing module for c++ inference

Kevindurant111:
Hi, guys. Recently, I have been working on developing C++ routines for Wenet. I have completed the development work, and the accuracy is aligned with your Python examples. This C++ post-processing module is based on Slyne's work, and it mainly involves some interface refactoring. This means that you no longer need to recompile third-party libraries like fst and KenLM when using these post-processing interfaces. Additionally, I have provided some utility functions, such as ctc_decoding and file reading functions. By the way, I am also developing a speech pre-processing module based on Armadillo, which is mainly used to extract fbank or mfcc features from WAV files. The reason for this is that Kaldi has too many dependencies, which is not very user-friendly for C++ users. I will gradually improve this module, and it will be open-sourced soon. I hope these are useful to you.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,6 @@
 [submodule "runtime/gpu/tensorrt_fastertransformer/FasterTransformer"]
 	path = runtime/gpu/tensorrt_fastertransformer/FasterTransformer
 	url = https://github.com/NVIDIA/FasterTransformer.git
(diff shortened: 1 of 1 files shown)


Kevindurant111:
@robin1001 @yuekaizhang

PR #1824, opened 2023-04-19
[bestrq] stable training

Mddct:
changes:
 -  stack fbank instead of stack features after subsampling  （BEST-RQ paper do stack fbank ）
 -  replace compute_mask_indices_v2 with compute_mask_indices,(v2 from wav2vec2  which compute mask in cpu), The v2 version can consider padding, and the probability of short sequence overmask becomes smaller
 - fbank should do cmvn ， stack fbanks should renorm using unlearnable LN  avoiding codebooks collapse
 - add fbank l2 regularization which can stable training
 - other info about training
 - reset encoder parameter using bert-style:  trunk_norm std==0.02

### DIFF
--- a/wenet/ssl/bestrq/bestqr_model.py
+++ b/wenet/ssl/bestrq/bestqr_model.py
@@ -1,8 +1,54 @@
 from typing import Optional, Tuple
 import torch
 
-from wenet.ssl.bestrq.mask import compute_mask_indices
+from wenet.ssl.bestrq.mask import compute_mask_indices_v2
 from wenet.utils.mask import make_pad_mask
+
+
+def quantize_vector(latent: torch.Tensor, codebook: torch.Tensor):
+    """
+    Symbols in comments:
+    B: batch_size.
+    D: latent_dim.
+    C: num_latent_classes per group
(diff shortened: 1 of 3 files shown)


Mddct:
BTW: ink error: ./test/test_tokenize.py:108:16: C419 Unnecessary list comprehension passed to all() prevents short-circuiting - rewrite as a generator. '

wenet/ssl/bestrq/bestqr_model.py
```diff
@@ -62,93 +174,106 @@ def forward(
...
-        # 2 mask features
-        # 2.0 apply mask
-        masked_xs, masked_masks = self._apply_mask(xs)
+        # force global cmvn
+        xs = xs - self.signal_mean
+        if self.signal_norm_var:
+            xs = xs * self.signal_istd
+        input = xs
+
+        features_pen: Optional[torch.Tensor] = None
```
xxchauncey, line 183: hi, I guess `xs` here is just the log mel fbank feature from dataloader, so it seems that calculating features_pen is not meaningful in this case.

PR #1898, opened 2023-07-03
Modify wenet.utils.cmvn

BenBill2077:
In the _load_kaldi_cmvn function, mudule logging and sys have not imported, this PR fix this bug.


robin1001:
too  many conflicts, please fix it.

PR #1921, opened 2023-07-21
[runtime/fix] fix locale wstring conversion error

robin1001:

PR #1922, opened 2023-07-21
fix(utils/string): add try-catch for wstring_converter

xingchensong:
Fix issue https://github.com/wenet-e2e/wenet/issues/745


xingchensong:
duplicate with https://github.com/wenet-e2e/wenet/pull/1921, closed

PR #1940, opened 2023-08-11
[RNN-T] transducer's simple_lm_proj shoud be predictor.output_size and delay_panalty is 0 when step less than 2*warmup_step

Zth9730:
fix some error
1. when use k2, predictor's output_size shoud return
2. about delay_penalty


Mddct:
cool

Mddct (approved):

PR #1944, opened 2023-08-12
[fix](paraformer): fix decoding bug for paraformer

xingchensong:
my result:
image

readme result:
image


robin1001 (approved):

robin1001:
Seems there is a big gap.

PR #2030, opened 2023-09-28
[wenet] add dropout for subsampling.py

Mddct:


robin1001:
Any experiment result?

robin1001:
It's better if we keep the `linear` unchanged, and just add dropout in the forward pass so that we can load the old pretrained models.

Mddct:
keep draft now until some results

PR #1003, opened 2022-03-31
self-supervised pretraining(wa2vec 2.0/data2vec) for wenet

Emiyassstar:
1.support self-supervised pretraining using wa2vec 2.0/data2vec method
截屏2021-11-16 上午11 36 43

PR #2034, opened 2023-10-07
[ssl/wav2vec2] support wav2vec2 

Mddct:
act subsampling as  feature extraction for wav2vec training  
plz ref  to https://github.com/wenet-e2e/wenet/pull/1003 for more info


Mddct:
The error in UnitTest was not involved  by this PR

PR #2058, opened 2023-10-17
[k2] move all k2 related functions to k2

robin1001:


xingchensong:
This is a follow-up PR from:
- https://github.com/wenet-e2e/wenet/pull/2057
- https://github.com/wenet-e2e/wenet/pull/2056

PR #2084, opened 2023-10-27
[cli] add doc

robin1001:


robin1001:
The lint error is caused by `flake8 --version` error.

PR #2101, opened 2023-11-02
support gpu for cli

yuekaizhang:
#2069 

Support gpu for cli. For jit cuda model, we may need several rounds of warmup to get acceleration.

### DIFF
--- a/wenet/cli/transcribe.py
+++ b/wenet/cli/transcribe.py
@@ -33,11 +33,11 @@ def get_args():
                         '--model_dir',
                         default=None,
                         help='specify your own model dir')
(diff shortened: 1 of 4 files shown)


wenet/cli/transcribe.py
```diff
@@ -33,6 +33,11 @@ def get_args():
                         '--model_dir',
                         default=None,
                         help='specify your own model dir')
+    parser.add_argument('-d',
+                        '--device',
```
robin1001, line 37: --gpu is better. we can use default -1 for cpu, and 0..N for gpu id.

robin1001 (approved):

PR #2158, opened 2023-11-23
fix bug

manbaaaa:
When printing the train loss without multiplying it by accum_grad, it results in the printed output of loss, loss_att, and loss_ctc being in different orders of magnitude.


xingchensong (approved):

xingchensong:
thx！

PR #2169, opened 2023-11-27
fix bug

manbaaaa:
1. Fix the issue of inconsistent magnitude between train loss and cv loss after setting the accum_grad parameter.
2. Restore the recording of train loss in TensorBoard to be consistent with the previous version.
3. Fix the calculation of Word Error Rate (WER) by replacing the "▁" symbol with a space when using BPE tokenization.


xingchensong:
下划线不能在这里去掉，对于使用了bpe的模型，要过完bpe decode后再去掉。例如librispeech https://github.com/wenet-e2e/wenet/blob/main/examples/librispeech/s0/run.sh#L217-L218

xingchensong:
https://github.com/wenet-e2e/wenet/issues/2170， 出于一致性考虑，建议tensorboard还是乘一下accum_grad

manbaaaa:
ok，I will revert the modification.

PR #2174, opened 2023-11-27
refactor(model): simplify __init__()

xingchensong:
简化 __init__ 的逻辑，避免大量 if-else

### TODO:
- [x] transformer / conformer
- [x] branchformer / efficient_conformer


robin1001:
个人觉得 wenet/utils 中更合适，可以把所有需要做 class mapping 的放在一起单独放一个新文件中。

wenet/utils/class_utils.py
```diff
@@ -0,0 +1,74 @@
...
+    assert rnn_type in ["rnn", "lstm", "gru"]
+    if rnn_type == "rnn":
+        return torch.nn.RNN
+    elif rnn_type == "lstm":
+        return torch.nn.LSTM
+    else:
+        return torch.nn.GRU
+
+
+WENET_SUBSAMPLE_CLASSES = {
```
robin1001, line 51: do you think it's more clear if we follow the name patter like before, such as `get_subsamling`, `get_embedding`, and `get_attention`.

xingchensong: `get_activation` return instance of class while `WENET_XXX_CLASSES` return class. To keep the name pattern, it's better to refine `get_actiavtion` to `WENET_ACTIVATION_CLASSES`, because we can not get  instance of subsampling class without arguments

robin1001: ok

robin1001 (approved):

PR #2292, opened 2024-01-10
[dataset] support speaker in dataset

robin1001:


xingchensong:
only support raw-mode?

xingchensong (approved):

PR #2330, opened 2024-02-01
Fix eot

Qiaochu-Song:
There is a bug when generating train.yaml from whisper checkpoint. Eot was using the token index of sot, and this pr fixes that.

### DIFF
--- a/examples/aishell/whisper/conf/finetune_whisper_largev3.yaml
+++ b/examples/aishell/whisper/conf/finetune_whisper_largev3.yaml
@@ -42,7 +42,7 @@ tokenizer_conf:
   non_lang_syms_path: null
   num_languages: 100
   special_tokens:
-    eot: 50258
+    eot: 50257
     no_speech: 50363
     no_timestamps: 50364
     sot: 50258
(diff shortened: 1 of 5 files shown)


xingchensong:
good catch! How did you find it? Does it affect trainining results?

xingchensong (approved):

Qiaochu-Song:
> good catch! How did you find it? Does it affect trainining results?

I was trying to convert and load whisper into wenet, then tried some samples without any fine-tuning. I was expecting it to give some reasonable outputs. However it did not stop decoding, even after <eot> was produced. Other than that the transcription looks correct. After changing eot to the correct idx, wenet decoding functions is able to terminate decoding correctly. 

Regarding effects on finetuning, I have not test it yet, but I assume it might affect training results a bit, but should not be too much, as tokens after <eot> won't be penalized in training.

xingchensong:
I see, THX !!!

xingchensong:
I can confirm that this fix does not affect training result.

- green: eot is 50258
- red: eot is 50257

image

Issue #2329, opened 2024-01-31
Blank token id hard coded in C++ decoders


zhr1201:
**Describe the bug**
Following the PR that adds whisper feature extraction https://github.com/wenet-e2e/wenet/pull/2320. The transcript produced in that PR contains lots of <|notimestampes|>.
This is because i exported whisper tokenizers with 99 languages, but for whisper large v3, tokenizer it's 100 langs. And the token corresponding to the extra output in the transcript is actually <|nospeech|>, which is used as the blank token in CTC. Fixing the token list does produce reasonable output, but still significantly worse than the one produced by transcribe.py that simulates a streaming inference.

I think there is still some mismatch in the decoder part, at least It's not possible to pass in blank token id through CLI.
For ctc prefix beam search, there is a field in the config https://github.com/wenet-e2e/wenet/blob/4c81459f9bf12bc0d9b2c2381b3d003c764bdb1a/runtime/core/decoder/ctc_prefix_beam_search.h#L30, but not passed in through CLI https://github.com/wenet-e2e/wenet/blob/4c81459f9bf12bc0d9b2c2381b3d003c764bdb1a/runtime/core/decoder/params.h#L75

and for wfst decoder, it assume the first token is blank. https://github.com/wenet-e2e/wenet/blob/4c81459f9bf12bc0d9b2c2381b3d003c764bdb1a/runtime/core/decoder/ctc_wfst_beam_search.cc#L80

**To Reproduce**
As described above

**Expected behavior**
CLI can handle blank tokens for different models, not just the case where the first token is blank.

[Bonus] decoder main produces the same result as transcribe.py for streaming finetuned whisper trained with hybrid ctc+attention loss

**Screenshots**

**Desktop (please complete the following information):**

**Smartphone (please complete the following information):**

**Additional context**

robin1001:
Yes, it's by design and it works fine before. It is a problem for whisper pretrained model since there is no blank in whisper tokenizer.

robin1001:
@xingchensong any comment?

xingchensong:
we should make it configurable like this :
- https://github.com/wenet-e2e/wenet/pull/2193
- https://github.com/wenet-e2e/wenet/pull/2157

PR #2366, opened 2024-02-22
[runtime] Configurable blank token idx

zhr1201:
Following https://github.com/wenet-e2e/wenet/pull/2320 which makes feature extraction pipeline compatible with whisper. We realized blank token idx is hard coded https://github.com/wenet-e2e/wenet/issues/2329, resulting in weird decoding results in the previous PR.

This PR is aiming at fixing that. **There might be some places that i missed, can you folks help me do another pass?**

Screenshot 2024-02-22 at 2 19 12 PM

Current RTF is 2.xx after warming up,  using one core with beam size of 8, on my local macbook. I think the model is probably capable of running at real time with avx512 on multiple cores, even with whisper large. so let's get it to work!

However, there is still some loose ends. Maybe you guys already know the answer and can share some insights.

- decoder main result **still slightly differs from transcribe.py** imitating streaming inference using attention masks. I don't know exactly why, maybe some default params are different?
- In order to create the decoding **TLG graph, some tools might needs to be updated to support flexible blank token id**. I haven't checked this one, but now this is not a high priority for us.

### DIFF
--- a/runtime/core/utils/string.cc
+++ b/runtime/core/utils/string.cc
@@ -153,7 +153,7 @@ std::string ProcessBlank(const std::string& str, bool lowercase) {
       std::wstring_convert<std::codecvt_utf8<wchar_t>, wchar_t> converter;
       std::wstring wsresult = converter.from_bytes(result);
       for (auto& c : wsresult) {
-        c = lowercase ? tolower(c, loc) : toupper(c, loc);
+        c = lowercase ? tolower(c, loc) : c;
       }
       result = converter.to_bytes(wsresult);
     } catch (std::exception& e) {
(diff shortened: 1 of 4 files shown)


runtime/core/utils/string.cc
```diff
@@ -153,7 +153,7 @@ std::string ProcessBlank(const std::string& str, bool lowercase) {
       std::wstring_convert<std::codecvt_utf8<wchar_t>, wchar_t> converter;
       std::wstring wsresult = converter.from_bytes(result);
       for (auto& c : wsresult) {
-        c = lowercase ? tolower(c, loc) : toupper(c, loc);
+        c = lowercase ? tolower(c, loc) : c;
```
zhr1201, line 156: this changes the --nolowercase behavior, we could also pass another flag if that's what we want to do

robin1001 (approved):

robin1001:
Great job!

PR #2407, opened 2024-03-12
fix mask bias dtype in sdpa

Mddct:


wenet/transformer/decoder.py
```diff
@@ -175,7 +175,7 @@ def forward(
         # tgt_mask: (B, L, L)
         tgt_mask = tgt_mask & m
         if self.use_sdpa:
-            tgt_mask = mask_to_bias(tgt_mask, tgt.dtype)
+            tgt_mask = mask_to_bias(tgt_mask, memory_mask.dtype)
             memory_mask = mask_to_bias(memory_mask, memory_mask.dtype)
```
xingchensong, line 180: should be `memory.dtype`, according to  https://github.com/wenet-e2e/wenet/blob/main/wenet/transformer/search.py#L293

xingchensong: both for `tgt_mask` and `memory_mask`

xingchensong (approved):

wenet/transformer/decoder.py
Mddct: ok 来回copy，copy错了

PR #2494, opened 2024-04-18
Using module kaldi_native_fbank, add the CPU feature extraction option.

76176235:
Using module kaldi_native_fbank, add the CPU feature extraction option.


76176235:
Code indentation error。

PR #2509, opened 2024-04-30
[Dataset] Resolving the 'Illegal seek' error

srdfjy:
When handling network resources, it is necessary to open tar files using the mode 'r|' instead of 'r:', as the former does not attempt seek operations and is suitable for sequential reading.

### DIFF
--- a/wenet/dataset/datapipes.py
+++ b/wenet/dataset/datapipes.py
@@ -337,7 +337,7 @@ class TarsDataPipe(IterDataPipe):
             assert 'stream' in sample
             try:
                 with tarfile.open(fileobj=sample['stream'],
-                                  mode="r:*") as stream:
+                                  mode="r|*") as stream:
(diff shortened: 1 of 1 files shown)


xingchensong:
when will illegal seek happen？

xingchensong:
seems there has a conflict https://github.com/wenet-e2e/wenet/pull/2301#issuecomment-1893027908

srdfjy:
When using shard mode, an "illegal seek" error can occur.

xingchensong:
> When using shard mode, an "illegal seek" error can occur.

I use shard for all my exps, and i never meet this issue

srdfjy:
> > When using shard mode, an "illegal seek" error can occur.
> 
> I use shard for all my exps, and i never meet this issue

This error occurs when reading data using HTTP+Shard.

Issue #2544, opened 2024-05-29
jit export transducer


Yonnie1331:
**Describe the bug**
Error when wenet/bin/export_jit.py a trained Transducer

**Additional context**
I think adding
`@torch.jit.unused`
before line 93 in wenet/transducer/transducer.py, def forward() will do good

Mddct:
pr welcome

Yonnie1331:
created! a lot thanks

PR #2548, opened 2024-06-01
unuse forward when jit exporting transducer

Yonnie1331:
Error when jit exporting transducer with previous code
Found this forward func is not necessary for jit export (like transformer/asr_model.py, line 77 and 78)


Mddct (approved):

Mddct:
https://github.com/wenet-e2e/wenet/issues/2544

PR #2554, opened 2024-06-13
Update subsampling.py

xu-gaopeng:
The StackNFramesSubsampling module was missing the pos_enc attribute, which caused an AttributeError when pos_enc.position_encoding was called in the forward_chunk.


Mddct (approved):

Mddct:
great

PR #2594, opened 2024-08-07
fix whisper tokenizer ut

Mddct:


Mddct:
resolved by https://github.com/wenet-e2e/wenet/pull/2593

PR #2614, opened 2024-08-16
[ssl/bestrq] happy ending---stable training 

Mddct:


xingchensong (approved):

ncakhoa:
@Mddct Can you explain when you stack fbank feature, why you change from normalizing freq dimension to time dimension and please provide some training results for comparing two approaches. I think normalizing freq dimension is more correct than time dimension, since the model wants to learn local distributions.

Mddct:
Thank you for your attention

截屏2024-08-19 11 55 32

The normalization of input is just to get richer codes id. The stack input dim=T is normalized here to simulate the calculation method of cmvn of stack input, and experiments show that this normalization will be better than norm. d has a larger proportion of unique codes

Mddct:
> @Mddct Can you explain when you stack fbank feature, why you change from normalizing freq dimension to time dimension and please provide some training results for comparing two approaches. I think normalizing freq dimension is more correct than time dimension, since the model wants to learn local distributions.



> @Mddct Can you explain when you stack fbank feature, why you change from normalizing freq dimension to time dimension and please provide some training results for comparing two approaches. I think normalizing freq dimension is more correct than time dimension, since the model wants to learn local distributions.

Thank you for your attention

截屏2024-08-19 11 55 32

The normalization of input is just to get richer codes id. The stack input dim=T is normalized here to simulate the calculation method of cmvn of stack input, and experiments show that this normalization will be better than norm. d has a larger proportion of unique codes

ncakhoa:
Can you show proportion of unique codes in two approaches. Also, have you tried to downstream model.

Mddct:
> Can you show proportion of unique codes in two approaches. Also, have you tried to downstream model.

Yes, some promising results will be released recently.

PR #2704, opened 2025-03-03
streaming batch chunk decoding method is available.

AdilAdam:
in original code, only single wav (bath==1) can be decodable while streaming chunk, it can be obtained that it is kind of not that fast while batch==1. In this update, multiple wavs can be inferencing, on the other hand, Batch size can be dynamic while origin batch size == 1,  (assert xs.size(0) == 1);


Mddct:
please fix lint

PR #2818, opened 2026-07-10
docs: add Tesla T4/Turing hardware notes (device, dtype, use_sdpa)

moduvoice:
## Motivation

Ran WeNet's pip-installed CLI/API on a real Tesla T4 (16GB, Turing/sm_75) and found several GPU-relevant behaviors that aren't documented anywhere, none of which involve any code or config changes:

- The README's headline quickstart (`wenet -m paraformer audio.wav`) never mentions GPU/device, and `--device` (CLI) / `device` (`wenet.load_model`) both default to `'cpu'` with no warning — on a GPU host this silently runs on CPU unless `--device cuda` is passed explicitly.
- `setup.py` pins `torch>=1.13.0` with no upper bound, so `pip install git+...` can pull a torch build newer than an older GPU driver supports, leaving `torch.cuda.is_available() == False` with no obvious error message tying it back to the install step.
- The pip-installed `wenet` CLI has no dtype flag at all (no `--dtype`/`--fp16`/`--bf16`), unlike the separate training-recipe script `wenet/bin/recognize.py`, which exposes `--dtype {fp16,fp32,bf16}`.
- On WeNet's own Conformer models (e.g. `wenetspeech`), the shipped `train.yaml` does not set `use_sdpa`, so it loads as `False` (manual matmul+softmax attention) by default. Enabling `use_sdpa: true` (same checkpoint, no retraining) measurably recovers most of the T4 bf16 slowdown.

## Changes

Doc-only, no code changes:

- `README.md`: note that the unpinned pip install can silently mismatch older drivers, with a verification command and a known-good reinstall example; note that the CLI/API run on CPU by default and how to opt into `--device cuda`.
- `docs/python_package.md`: new "GPU/Hardware Notes" section documenting, with numbers measured on a Tesla T4:
  - CUDA vs CPU: **2.84x** faster (0.205s vs 0.581s per utterance, cached paraformer model)
  - No dtype flag on the pip CLI vs. `wenet/bin/recognize.py`'s `--dtype {fp16,fp32,bf16}`
  - bf16 cast on WeNet Conformer: **~47% slower** than fp32 (46.6ms vs 31.7ms) — Turing has no bf16 tensor-core acceleration; fp16 shows no speedup at batch size 1 but cuts VRAM ~46%
  - `use_sdpa: true` (not set in shipped configs) sped things up 7-29% with identical output text: fp32 31.7ms→29.4ms, fp16 32.2ms→30.0ms, bf16 46.6ms→33.1ms — this only applies to WeNet's own Conformer models, not Paraformer (SANM attention has no `use_sdpa` path)
  - int8 dynamic quantization (`torch.quantization.quantize_dynamic`) is CPU-only; raises `NotImplementedError` on CUDA (PyTorch backend limitation, not a WeNet bug)

## Testing

- Verified all measurements above directly against this repo's public API (`wenet.load_model`/`model.transcribe`, `wenet.utils.init_model.init_model`+`model.decode`) on a real Tesla T4 16GB (driver 550.163.01, torch 2.4.0+cu121).
- `use_sdpa` toggle was validated by re-running the same checkpoint with `encoder_conf['use_sdpa']=True`/`decoder_conf['use_sdpa']=True` and confirming output text is byte-identical across all 8 dtype/use_sdpa combinations tested.
- No source files under `wenet/` were modified — only `README.md` and `docs/python_package.md`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,6 +29,22 @@
 pip install git+https://github.com/wenet-e2e/wenet.git
 ```
 
+This pulls in `torch>=1.13.0` with no upper bound, so pip may install a torch
+build newer than your installed GPU driver supports. If you plan to use a GPU,
+verify it after install:
+
+``` sh
+python -c "import torch; print(torch.cuda.is_available())"
+```
+
+If this prints `False` on a machine with a GPU, torch silently falls back to
+CPU with no error. Reinstall a torch build matching your driver's CUDA
+version, e.g.:
+
+``` sh
+pip install torch==2.4.0+cu121 torchaudio==2.4.0+cu121 --index-url https://download.pytorch.org/whl/cu121 --force-reinstall
+```
+
 **Command-line usage** (use `-h` for parameters):
 
 ``` sh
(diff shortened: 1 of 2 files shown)


robin1001:
Great!

